Issue #32, opened 2022-08-26
Support not only 'latest' releases


chshersh:
Currently `tool-sync` downloads the latest release by default. I can imagine some people preferring specific releases. This should be easily supported by `tool-sync` with a few code changes:

- [ ] Add the `tag` field to `ConfigAsset`
    * https://github.com/chshersh/tool-sync/blob/e908503d504b8122459300ed8b4092a9fd63e5a2/src/config/schema.rs#L22-L37
- [ ] Patch TOML decoding
    * https://github.com/chshersh/tool-sync/blob/e908503d504b8122459300ed8b4092a9fd63e5a2/src/config/toml.rs#L60-L72
- [ ] Update the `ToolInfo` to store a value of a new enum with constructors `Latest` or `Tag`
    * https://github.com/chshersh/tool-sync/blob/e908503d504b8122459300ed8b4092a9fd63e5a2/src/model/tool.rs#L29-L43
- [ ] Patch `configure.rs` to apply `ConfigAsset` settings
    * https://github.com/chshersh/tool-sync/blob/main/src/sync/configure.rs
- [ ] Update `release_url` to return not only the latest release
    * https://github.com/chshersh/tool-sync/blob/e908503d504b8122459300ed8b4092a9fd63e5a2/src/sync/download.rs#L26-L32
- [ ] Add tests

PR #60, opened 2022-09-02
Resolves #49

### LINKED ISSUE #49: Specify which version of the tool was actually downloaded
This could be a helpful improvement in the output. I imagine the output to be of the form:

```haskell
✅  rg  (v13.0.0) Completed!
✅  exa (v0.10.1) Completed!
```

To do this:

- [ ] Add `tag_name` to the `Release` type here
    * https://github.com/chshersh/tool-sync/blob/0afde4fa01e95910b25397fa55898527a9f24724/src/model/release.rs#L3-L7
- [ ] Patfh `prefix` formatting to specify the tool version as well
    * https://github.com/chshersh/tool-sync/blob/0afde4fa01e95910b25397fa55898527a9f24724/src/sync/progress.rs#L28-L32

This may require to add more formatting options:

https://github.com/chshersh/tool-sync/blob/cb339b09238964e8dc2d1a72389b6624c5bb7599/src/sync/progress.rs#L4-L7

MitchellBerend:
This resolves #49

I took a stab at implementing this but Im not sure if my approach is the right
way to go. Because the list of tools is already known before actually calling
the github api the length of these names is also known. This is not that case
with the `tag_name`. The tags from the example file are all less than 7 long but
that does not guarantee other repos.

I could try and check what tags are available for all tools before actually
downloading them. Can I get your input on this?

```bash
$ cargo run -- --config tools.toml sync
    Finished dev [unoptimized + debuginfo] target(s) in 0.02s
     Running `target/debug/tool --config tools.toml sync`
✅  bat        v0.21.0  Completed!
✅  difftastic 0.34.0   Completed!
✅  exa        v0.10.1  Completed!
✅  fd         v8.4.0   Completed!
✅  ripgrep    13.0.0   Completed!
```

### DIFF
--- a/src/model/release.rs
+++ b/src/model/release.rs
@@ -2,7 +2,7 @@ use serde::Deserialize;
 
 #[derive(Deserialize, Debug)]
 pub struct Release {
-    // tag_name: String,
+    pub tag_name: String,
     pub assets: Vec<Asset>,
 }
 
--- a/src/sync/download.rs
+++ b/src/sync/download.rs
@@ -20,6 +20,7 @@ pub struct Downloader<'a> {
 pub struct DownloadInfo {
     pub archive_path: PathBuf,
     pub asset_name: String,
+    pub tag_name: String,
 }
 
 impl<'a> Downloader<'a> {
@@ -106,6 +107,7 @@ impl<'a> Downloader<'a> {
                 Ok(DownloadInfo {
                     archive_path,
                     asset_name: asset.name.clone(),
+                    tag_name: release.tag_name,
                 })
             }
         }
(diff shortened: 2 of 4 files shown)


chshersh:
@MitchellBerend Wow, thanks a lot for working on this! I'll have a look as soon as I can 🤗

MitchellBerend:
#61 might actually solve the issue that I have with my pr since it does actually give some idea about the length of the tag before the program reaches out to github. We can maybe assume `latest` as the lower bound and the longest tag as the upper bound like you did with `max_tool_size`

chshersh (approved):
@MitchellBerend Thanks a lot for the implementation!

I think this change is good on its own so I'm happy to merge it 👏🏻 
Further improvements can be done after #32 is fixed (for example, it would be nice to see the version during the `Downloading` phase as well). But I propose to do this after #61 merged 🚢

PR #62, opened 2022-09-03
Adds a generate command for a default .tools.toml file

### LINKED ISSUE #52: Add a command to output default config with comments and explanations
It can be helpful to get the config without the need to look it up externally.

MitchellBerend:
Resolves #52.

The generate command also uses the --config flag to set the location of the generated file. Would it be a good idea to have some output when actually generating the config? Right now there is no actual response unless there is some file system error.

```bash
$ cargo run -- --config test-tools.toml generate
    Finished dev [unoptimized + debuginfo] target(s) in 2.14s
     Running `target/debug/tool --config test-tools.toml generate``

$ cat test-tools.toml 
#
#store_directory = "$HOME/.tools.toml"
#
#    [bat]
#        owner =     "sharkdp"
#        repo =      "bat"
#        exe_name =  "bat"
#        tag =       "latest"
#    [exa]
#        owner =     "ogham"
#        repo =      "exa"
#        exe_name =  "exa"
#        tag =       "latest"
#    [fd]
#        owner =     "sharkdp"
#        repo =      "fd"
#        exe_name =  "fd"
#        tag =       "latest"
#    [ripgrep]
#        owner =     "BurntSushi"
#        repo =      "ripgrep"
#        exe_name =  "rg"
#        tag =       "latest"
```


MitchellBerend:
We should wait for the discussion in #65 to be resolved since this issue might constitute changes in this pr

MitchellBerend:
The file that currently gets generated looks like this

```bash
$ cat test-tools.toml 
#
# The directory to store all tools in
#store_directory = "$HOME/.local/bin"
#
#[ripgrep]
#   owner     = "BurntSushi"
#   repo      = "ripgrep"
#   exe_name  = "rg"
#
#   # Uncomment to download a specific version or tag.
#   # Without this tag latest will be used
#   # tag       = "latest"
#
#
# Asset name to download on linux OSes
#asset_name.linux = "x86_64-unknown-linux-musl"
#
# uncomment if you want to install on macOS as well
#asset_name.macos = "apple-darwin"
#
# uncomment if you want to install on Windows as well
#asset_name.windows = "x86_64-pc-windows-msvc"
```

Is this missing anything?

src/lib.rs
```diff
@@ -46,3 +57,10 @@ fn resolve_config_path(config_path: Option<PathBuf>) -> PathBuf {
         },
     }
 }
+
+fn generate_config(config_path: &PathBuf) -> Result<&PathBuf, std::io::Error> {
+    let prepare_path = Some(config_path.to_owned());
+    let path = resolve_config_path(prepare_path);
+    std::fs::write(path, config_template::CONFIG_TEMPLATE)?;
+    Ok(config_path)
```
chshersh, line 65: I'm a bit unsatisfied by this behaviour. According the [CLI Guidelines](https://clig.dev/#output), CLI tools shouldn't create/update files if the user didn't specify the path explicitly (unless those files store internal state of a CLI tool but this is not the case with the config file).

So, with the current design, when a user runs `tool default-config`, the command would unrecoverably override `$HOME/.tool.toml` and this is undesirable.

I believe, the following behaviour (although more complicated) is better:

+ **(required minimum**) `tool default-config` should only print the config to stdout. If a user wants to save the config to a file, they can just do `tool default-config > path/to/config`.
    > I'm thinking about adding the `--path` option to the `default-config` command, so in the future it should be even possible to write something like this:
    > ```shell
    > tool default-config > $(tool default-config --path)
    > ```
+ **(optional)** If the `--config` flag is specified, the default configuration is saved in the explicitly specified path. If this file already exist, ask interactively whether user want's to change it. And if the answer is "yes" then override it. The `defaulg-config` command can also accept the `-f|--force` flag to ignore interactive question.
+ **(alternative 1)** Move the `--config` flag inside the `sync` command and don't bother about the the `default-config` behaviour. This is a UX breaking change but, oh, well...
+ **(alternative 2)** Always require the `--config` flag with the `default-config` command

What do you think?

src/config_template.rs
```diff
@@ -0,0 +1,26 @@
...
+#store_directory = "$HOME/.local/bin"
+#
+#[ripgrep]
+#   owner     = "BurntSushi"
+#   repo      = "ripgrep"
+#   exe_name  = "rg"
+#
+#   # Uncomment to download a specific version or tag.
+#   # Without this tag latest will be used
+#   # tag       = "latest"
```
chshersh, line 15: I think that `tag` field should be `13.0.0` because `latest` is not supported

```suggestion
#   # tag       = "13.0.0"
```

src/config_template.rs
```diff
@@ -0,0 +1,26 @@
...
+#
+#
+# Asset name to download on linux OSes
+#asset_name.linux = "x86_64-unknown-linux-musl"
+#
+# uncomment if you want to install on macOS as well
+#asset_name.macos = "apple-darwin"
+#
+# uncomment if you want to install on Windows as well
+#asset_name.windows = "x86_64-pc-windows-msvc"
```
chshersh, line 25: Let's indent all `asset.*` fields to match the `owner/...` fields indentation. TOML is not layout sensitive but still, it could be good visually to say that they belong to the same section.

src/config_template.rs
```diff
@@ -0,0 +1,26 @@
+/// This file only holds the template that is used to generate a default .tools.toml.
+
+pub const CONFIG_TEMPLATE: &str = r##"# This file was automatically generated by tool-sync
```
chshersh, line 3: Let's add a header with more info:

```
# tool-sync default configuration file
####################################################
```

> Would be cool to also say by which version of `tool-sync` but not sure how to do this without hardcoding the version 🤔 

I'm taking the inspiration from the `hlint --default` command (Haskell linter)

<details>
    <summary> HLint default </summary>

```
# HLint configuration file
# https://github.com/ndmitchell/hlint
##########################

# This file contains a template configuration file, which is typically
# placed as .hlint.yaml in the root of your project


# Specify additional command line arguments
#
# - arguments: [--color, --cpp-simple, -XQuasiQuotes]


# Control which extensions/flags/modules/functions can be used
#
# - extensions:
#   - default: false # all extension are banned by default
#   - name: [PatternGuards, ViewPatterns] # only these listed extensions can be used
#   - {name: CPP, within: CrossPlatform} # CPP can only be used in a given module
#
# - flags:
#   - {name: -w, within: []} # -w is allowed nowhere
#
# - modules:
#   - {name: [Data.Set, Data.HashSet], as: Set} # if you import Data.Set qualified, it must be as 'Set'
#   - {name: Control.Arrow, within: []} # Certain modules are banned entirely
#
# - functions:
#   - {name: unsafePerformIO, within: []} # unsafePerformIO can only appear in no modules


# Add custom hints for this project
#
# Will suggest replacing "wibbleMany [myvar]" with "wibbleOne myvar"
# - error: {lhs: "wibbleMany [x]", rhs: wibbleOne x}

# The hints are named by the string they display in warning messages.
# For example, if you see a warning starting like
#
# Main.hs:116:51: Warning: Redundant ==
#
# You can refer to that hint with `{name: Redundant ==}` (see below).

# Turn on hints that are off by default
#
# Ban "module X(module X) where", to require a real export list
# - warn: {name: Use explicit module export list}
#
# Replace a $ b $ c with a . b $ c
# - group: {name: dollar, enabled: true}
#
# Generalise map to fmap, ++ to <>
# - group: {name: generalise, enabled: true}


# Ignore some builtin hints
# - ignore: {name: Use let}
# - ignore: {name: Use const, within: SpecialModule} # Only within certain modules


# Define some custom infix operators
# - fixity: infixr 3 ~^#^~


# To generate a suitable file for HLint do:
# $ hlint --default > .hlint.yaml

```
</details>

src/config_template.rs
```diff
@@ -0,0 +1,26 @@
...
+#
+# The directory to store all tools in
+#store_directory = "$HOME/.local/bin"
+#
```
chshersh, line 7: Let's also add a section by saying that `tool-sync` natively supports some common tools. Something like

```
#
# [bat]
# [difftastic]
# [fd]
```

chshersh (changes requested):
This is a great change 🏆 

I feel like the design space for this innocently-looking option is too big so I'd like to have a conversation. But good defaults matter 👏🏻 So I want users to have the best UX.

src/config_template.rs
MitchellBerend: We can do some metaprogramming in as `build.rs` by parsing the `Cargo.toml` and finding the version there. Im not sure how this works with ci or github releases though.

Maybe we can also use the automatically generated `--version` flag clap provides for convenience. This is probably a more clean solution but I can only find a reference to a `get_version` method on the `App` struct which is deprecated.

src/config/cli.rs
```diff
@@ -17,4 +17,7 @@ pub struct Cli {
 pub enum Command {
     /// Sync all tools specified in configuration file
     Sync,
+
+    /// Generate a default .tools.toml file
+    Generate,
```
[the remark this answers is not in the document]
MitchellBerend: Clap indeed converts it like that. Looking at the command now it is indeed very generic/non descriptive.

src/lib.rs
MitchellBerend: I like having the smallest feature as possible so, although I like the optional and alternative 2 ideas, I think those should be their own features.

README.md
```diff
@@ -109,13 +109,16 @@ the following command (Unix-only):
 ```shell
 curl https://raw.githubusercontent.com/chshersh/tool-sync/main/example-tool-sync-config.toml > ~/.tool.toml
 ```
+A default config can be also be generated by running `tool --config=path/to/config generate`.
+This will generate an example file at `path/to/config`.
```
chshersh, line 113: Let's fix the description as well 👍🏻

chshersh (commented):
Just one minor comment and this PR is ready to go 👍🏻 

Really appreciate your work here, @MitchellBerend 👏🏻 
The more I think about this, the more ideas and feature requests I can come up with. But let's take time and settle on some design and refine it later.

README.md
MitchellBerend: Ah I totally forgot about this one, good call.

chshersh (approved):
Thanks a ton!

chshersh:
@MitchellBerend These new issues look like a good improvement over the existing state of affairs 👍🏻 
I'm thinking about a completely alternative design for the config-related commands (e.g. maybe it would be better to have the `tool config --default` command and other flags to the `config` command?) so I'll create an RFC soon and I'd like to invite you to the discussion. And when it's decided what interface is better, we can continue with other improvements around configs 🚢

chshersh:
I've created an issue to discuss the future interface of all config-related commands:

* https://github.com/chshersh/tool-sync/issues/75

PR #64, opened 2022-09-03
[#63] Added tag support in download progress

### LINKED ISSUE #63: Adding tag support in download progress output
As discussed in #60 tag support is now actually possible from the start of the
process thanks to PR #61.

The max tag size can probably be passed to the formatter the same way the
`max_tool_size` is passed now. I can try and implement this but it might also
be a good first issue for others so they can familiarize themselves with the
code.

MitchellBerend:
Resolves #63

Added a `max_tag_size` on `SyncProgress` to prepare the output message for tags
that are longer than normal.

```bash
$ cargo run -- --config test-tools.toml sync
    Finished dev [unoptimized + debuginfo] target(s) in 0.04s
     Running `target/debug/tool --config test-tools.toml sync`
⛔  bat     asdkjhkasdjfaksjdfhk;j [error] https://api.github.com/repos/sharkdp/bat/releases/tags/asdkjhkasdjfaksjdfhk;j: status code 404
✅  exa     v0.10.1                Completed!
✅  fd      v8.4.0                 Completed!
✅  ripgrep 13.0.0                 Completed!


$ cat test-tools.toml 
#
store_directory = "./bins"

    [bat]
        owner =     "sharkdp"
        repo =      "bat"
        exe_name =  "bat"
        tag =       "asdkjhkasdjfaksjdfhk;j"
    [exa]
        owner =     "ogham"
        repo =      "exa"
        exe_name =  "exa"
        tag =       "v0.10.1"
    [fd]
        owner =     "sharkdp"
        repo =      "fd"
        exe_name =  "fd"
        tag =       "v8.4.0"
    [ripgrep]
        owner =     "BurntSushi"
        repo =      "ripgrep"
        exe_name =  "rg"
        tag =       "13.0.0"
```

### DIFF
--- a/src/sync.rs
+++ b/src/sync.rs
@@ -32,7 +32,12 @@ For more details, refer to the official documentation:
         let store_directory = config.ensure_store_directory();
 
         let tools: Vec<String> = config.tools.keys().cloned().collect();
-        let sync_progress = SyncProgress::new(tools);
+        let tags: Vec<String> = config
+            .tools
+            .values()
+            .map(|config_asset| { config_asset.tag.clone().unwrap_or_else(|| "latest".into()) })
+            .collect();
+        let sync_progress = SyncProgress::new(tools, tags);
         let installer = Installer::mk(store_directory, sync_progress);
 
         for (tool_name, config_asset) in config.tools.iter() {
--- a/src/sync.rs
+++ b/src/sync.rs
@@ -35,7 +35,7 @@ For more details, refer to the official documentation:
         let tags: Vec<String> = config
             .tools
             .values()
-            .map(|config_asset| { config_asset.tag.clone().unwrap_or_else(|| "latest".into()) })
+            .map(|config_asset| config_asset.tag.clone().unwrap_or_else(|| "latest".into()))
             .collect();
         let sync_progress = SyncProgress::new(tools, tags);
         let installer = Installer::mk(store_directory, sync_progress);
--- a/src/sync/progress.rs
+++ b/src/sync/progress.rs
@@ -19,7 +19,14 @@ impl SyncProgress {
         let max_tool_size = tools.iter().map(|tool| tool.len()).max().unwrap();
 
         // putting a default of 8 here since tags like v0.10.10 is already 8
-        let max_tag_size = tags.iter().map(|tag| tag.len()).max().unwrap_or(8);
+        let max_tag_size = tags
+            .iter()
+            .map(|tag| match tag.len() {
+                0..=8 => 8,
+                _ => tag.len(),
+            })
+            .max()
+            .unwrap_or(8);
 
         let multi_progress = MultiProgress::new();
 
--- a/src/sync/progress.rs
+++ b/src/sync/progress.rs
@@ -3,6 +3,7 @@ use indicatif::{MultiProgress, ProgressBar, ProgressStyle};
 
 pub struct SyncProgress {
     max_tool_size: usize,
+    max_tag_size: usize,
     multi_progress: MultiProgress,
 }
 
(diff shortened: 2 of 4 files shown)


src/sync.rs
```diff
@@ -32,7 +32,12 @@ For more details, refer to the official documentation:
         let store_directory = config.ensure_store_directory();
 
         let tools: Vec<String> = config.tools.keys().cloned().collect();
-        let sync_progress = SyncProgress::new(tools);
+        let tags: Vec<String> = config
+            .tools
+            .values()
+            .map(|config_asset| config_asset.tag.clone().unwrap_or_else(|| "latest".into()))
+            .collect();
+        let sync_progress = SyncProgress::new(tools, tags);
```
chshersh, line 40: It's a clever solution 👍🏻 
Indeed, we can assume that the version is `latest` if it's not specified in the config. So we can iterate through all of them before actually doing the processing.

Probably, in some future, we can improve this even further by fetching the release info about each tool first and producing the **real** version instead of `latest`. But that's a separate feature request 🙂

src/sync/progress.rs
```diff
@@ -13,36 +14,48 @@ const PROCESS: Emoji<'_, '_> = Emoji("📥  ", ".. ");
...
         // unwrap is safe here because 'new' is called with a non-empty vector
         let max_tool_size = tools.iter().map(|tool| tool.len()).max().unwrap();
 
+        // putting a default of 8 here since tags like v0.10.10 is already 8
+        let max_tag_size = tags
+            .iter()
+            .map(|tag| match tag.len() {
+                0..=8 => 8,
+                _ => tag.len(),
+            })
```
chshersh, line 27: I would simplify this match by using [std::cmp::max](https://doc.rust-lang.org/std/cmp/fn.max.html) function, so it could be just `max(tag.len(), 8)` instead of `match ...`.

Also, let's move 8 to a top-level `const` with the name like `MIN_TAG_LEN` and use it everywhere

src/sync/progress.rs
```diff
@@ -66,11 +79,58 @@ impl SyncProgress {
...
+        pb.set_prefix(self.fmt_prefix(FAILURE, tool_name, tag_name));
 
         let failure_msg = format!("{}", style(err_msg).red());
         pb.set_message(failure_msg);
         pb.finish();
     }
 }
+
+#[cfg(test)]
+mod tests {
```
chshersh, line 92: The tests are 👨🏻‍🍳 🤌

chshersh (changes requested):
Great and simple improvement 👍🏻 

I have a few suggestions for improving the code but the overall it looks great 👏🏻

src/sync/progress.rs
MitchellBerend: Good call actually on the const actually, this way there are stray numbers left in the code if the minimum size changes for whatever reason.

As for the map I think its more clear to use the absolute name as `max` is also called on the map itself after the collection is prepared. Is that acceptable or do you have a different suggestion?

```rust
        let max_tag_size = tags
            .iter()
            .map(|tag| std::cmp::max(tag.len(), MIN_TAG_SIZE) )
            .max()
            .unwrap_or(8);
```

chshersh: I'm okay with using `std::cmp::max`, that sounds reasonable to me 👍🏻 

Do you also want to change the `.unwrap_or(8);` to `.unwrap_or(MIN_TAG_SIZE);`? I usually don't like magic numbers and it's better to avoid them when possible. I sometimes might be lazy and use a constant but it's not sustainable so let's improve the code that we use 🏡

chshersh (approved):
Looks amazing 👏🏻

PR #74, opened 2022-09-09
added pull request template

MitchellBerend:
I thought it would be nice to have a pr template so I took inspiration from the kubernetes one. This aims to have a more consistent process for prs moving forward.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,86 @@
+<!-- Based on the Kubernetes pull request template -->
+
+
+<!--  Thanks for sending a pull request!  Here are some tips for you:
+
+1. If this is your first time, please read our contributor guidelines:
+https://github.com/chshersh/tool-sync/blob/main/CONTRIBUTING.md
+
+
+2. Please label this pull request according to what type of issue you are
+addressing, especially if this is a release targeted pull request. For reference
+on required PR/issue labels, read here:
+https://git.k8s.io/community/contributors/devel/sig-release/release.md#issuepr-kind-label
+
+
+3. Ensure you have added or ran the appropriate tests for your PR
+
+
+4. If the PR is unfinished mark it as draft
+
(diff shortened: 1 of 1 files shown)


chshersh:
@MitchellBerend This is a great idea 👏🏻 

My only feedback would be to simplify the template as `tool-sync` is not as complicated a project as K8S 😅 `tool-sync` also has a much smaller user base. So we can grow the number of contributors by lowering contributing requirements at this stage. In my vision, long PR templates can be off-putting when all people want is to add a patch.

A few remarks:

1. I don't think we need to mention `CONTRIBUTING.md` and we can rely on GitHub UI here that warns about reading `CONTRIBUTING.md`. It's also my general expectation that people read the contributing guide before contributing to the project.
2. The `/kind` commands can be removed. We don't have the tooling to automatically set labels based on this and we can set them manually after PR is open.
3. External contributors can't set the labels on their own.

Based on these remarks, I propose the following template:

```markdown
<--! PR title template: [#ISSUE_NUMBER] <short description>
     Example:           [#42] Solve problem X
-->

Resolves #PUT_ISSUE_NUMBER_HERE_IF_APPLICABLE

Write description of your changes here

### Additional tasks

- [ ] Tests added
```

MitchellBerend:
> long PR templates can be off-putting when all people want is to add a patch.

I can definitely understand that a long doc can be off putting for others even though I like longer docs.
I will add your template since I think we should have some template so there is some amount of similar structure in prs.

MitchellBerend:
Is there a way to skip ci for changes like this? I don't think it makes sense to waste resources when a template or `README.md` gets changed since we don't have doc tests (yet).

chshersh (approved):

chshersh:
@MitchellBerend You can add `[skip ci]` to the commit message and CI won't run for this commit. It can be handy for documentation-only changes

* https://github.blog/changelog/2021-02-08-github-actions-skip-pull-request-and-push-workflows-with-skip-ci/

Issue #55, opened 2022-08-30
Don't require 'exe_name'


chshersh:
Currently, `tool-sync` requires to specify `exe_name` in the configuration. However, it often happens that the tool has exactly the same name as the repository. In this case, we could allow to not specify the `exe_name` field and default it to the repo name.

To achieve this, we should patch the tool configuration in relevant places (and add tests as well):

Relevant code:

* https://github.com/chshersh/tool-sync/blob/cb339b09238964e8dc2d1a72389b6624c5bb7599/src/sync/configure.rs#L30-L46
* https://github.com/chshersh/tool-sync/blob/cb339b09238964e8dc2d1a72389b6624c5bb7599/src/sync/configure.rs#L60-L63

PR #76, opened 2022-09-09
[#55] fetch exe_name from api if ommited

mmohammadi9812:
Resolves [#55]

### DIFF
--- a/src/sync/configure.rs
+++ b/src/sync/configure.rs
@@ -2,7 +2,6 @@ use crate::config::schema::ConfigAsset;
 use crate::model::asset_name::AssetName;
 use crate::model::tool::{Tool, ToolError, ToolInfo, ToolInfoTag};
 use crate::sync::db::lookup_tool;
-use serde::Deserialize;
 
 pub fn configure_tool(tool_name: &str, config_asset: &ConfigAsset) -> Tool {
     match lookup_tool(tool_name) {
@@ -28,31 +27,14 @@ pub fn configure_tool(tool_name: &str, config_asset: &ConfigAsset) -> Tool {
     }
 }
 
-#[derive(Deserialize)]
-pub struct RepoInformation {
-    pub name: String,
-    // There are other fields too, but we don't need them now
-    // They may be added later if needed at all
-}
-
-fn get_exe_name(owner: String, repo: String) -> Result<String, ureq::Error> {
-    let url = format!(
-        "https://api.github.com/repos/{owner}/{repo}",
-        owner = owner,
-        repo = repo
-    );
-    let repo: RepoInformation = ureq::get(&url).call()?.into_json()?;
-    Ok(repo.name)
-}
-
 /// Configure 'ToolInfo' completely from 'ConfigAsset'
 fn full_configure(config_asset: &ConfigAsset) -> Option<ToolInfo> {
     let owner = config_asset.owner.clone()?;
     let repo = config_asset.repo.clone()?;
-    let exe_name = match config_asset.exe_name.clone() {
-        Some(exe_name) => exe_name,
-        None => get_exe_name(owner.clone(), repo.clone()).unwrap(),
-    };
+    let exe_name = config_asset
+        .exe_name
+        .clone()
+        .unwrap_or(config_asset.repo.clone()?);
     let tag = config_asset
         .tag
         .clone()
@@ -87,7 +69,7 @@ impl ToolInfo {
             exe_name: config_asset
                 .exe_name
                 .clone()
-                .unwrap_or_else(|| self.exe_name.clone()),
+                .unwrap_or(config_asset.repo.clone().unwrap()),
             asset_name: AssetName {
                 linux: config_asset
                     .asset_name
@@ -241,7 +223,7 @@ mod tests {
     }
 
     #[test]
-    fn get_exe_name() {
+    fn empty_exe_name() {
         let tool_name = "abcdef";
 
         let config_asset = ConfigAsset {
--- a/src/sync/configure.rs
+++ b/src/sync/configure.rs
@@ -2,6 +2,7 @@ use crate::config::schema::ConfigAsset;
 use crate::model::asset_name::AssetName;
 use crate::model::tool::{Tool, ToolError, ToolInfo, ToolInfoTag};
 use crate::sync::db::lookup_tool;
+use serde::Deserialize;
 
 pub fn configure_tool(tool_name: &str, config_asset: &ConfigAsset) -> Tool {
     match lookup_tool(tool_name) {
@@ -27,11 +28,31 @@ pub fn configure_tool(tool_name: &str, config_asset: &ConfigAsset) -> Tool {
     }
 }
 
+#[derive(Deserialize)]
+pub struct RepoInformation {
+    pub name: String,
+    // There are other fields too, but we don't need them now
+    // They may be added later if needed at all
+}
+
+fn get_exe_name(owner: String, repo: String) -> Result<String, ureq::Error> {
+    let url = format!(
+        "https://api.github.com/repos/{owner}/{repo}",
+        owner = owner,
+        repo = repo
+    );
+    let repo: RepoInformation = ureq::get(&url).call()?.into_json()?;
+    Ok(repo.name)
+}
+
 /// Configure 'ToolInfo' completely from 'ConfigAsset'
 fn full_configure(config_asset: &ConfigAsset) -> Option<ToolInfo> {
     let owner = config_asset.owner.clone()?;
     let repo = config_asset.repo.clone()?;
-    let exe_name = config_asset.exe_name.clone()?;
+    let exe_name = match config_asset.exe_name.clone() {
+        Some(exe_name) => exe_name,
+        None => get_exe_name(owner.clone(), repo.clone()).unwrap(),
+    };
     let tag = config_asset
         .tag
         .clone()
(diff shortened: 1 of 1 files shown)


chshersh (changes requested):
Thanks for the contribution!

It's a nice idea and the implementation looks rather small 🤔 Though, I think the approach of requesting this information from GitHub as a separate API request increases the cost of the operation significantly.

I believe the main code that requires changes is in the `sync/configure.rs` module:

https://github.com/chshersh/tool-sync/blob/cb339b09238964e8dc2d1a72389b6624c5bb7599/src/sync/configure.rs#L60-L63

Here it should take the value of `repo` if `exe_name` is `None`.

Previously, I was also thinking about changing the TOML parsing by specifying the `exe_name` as `repo` always (as the original issue says). But now I'm thinking that this can lead to potentially confusing behaviour when users override only `repo` name and suddenly their executable name changes as well.

So changes in `sync/configure.rs` is the only thing that is required here (in addition to tests and we can avoid querying the GitHub API 🤗 It's important because there's a pretty strict limit on it 😮‍💨

src/sync/configure.rs
```diff
@@ -27,11 +28,31 @@ pub fn configure_tool(tool_name: &str, config_asset: &ConfigAsset) -> Tool {
...
+    );
+    let repo: RepoInformation = ureq::get(&url).call()?.into_json()?;
+    Ok(repo.name)
+}
+
 /// Configure 'ToolInfo' completely from 'ConfigAsset'
 fn full_configure(config_asset: &ConfigAsset) -> Option<ToolInfo> {
     let owner = config_asset.owner.clone()?;
     let repo = config_asset.repo.clone()?;
-    let exe_name = config_asset.exe_name.clone()?;
```
MitchellBerend, line 34: If this line is just
```rust
    let exe_name = config_asset.exe_name.clone().unwrap_or(config_asset.repo.clone()?);
```
Can we not let it error and give some type of output to the user to add an `exe_name` to the config?
This way there is no need for an extra api call.

Another approach would be looking for the common substring in the asset names. But I am more in favor of letting the error bubble up so the tool does not download something if the user forgets to add an `exe_name` or they mistake the name of the binary with the repo.

MitchellBerend: Here is a small piece of the json returned by the repos/{owner}/repo/release/tags/{tag} endpoint.

The name is always formatted as `{name}-{tag}-{arch}.some.zip.extension` 

---

```json

  "assets": [
    {
      "url": "https://api.github.com/repos/BurntSushi/ripgrep/releases/assets/38486868",
      "id": 38486868,
      "node_id": "MDEyOlJlbGVhc2VBc3NldDM4NDg2ODY4",
      "name": "ripgrep-13.0.0-arm-unknown-linux-gnueabihf.tar.gz",
      "label": "",
     ...
      "content_type": "application/octet-stream",
      "state": "uploaded",
      "size": 1763861,
      "download_count": 3387,
      "created_at": "2021-06-12T12:31:57Z",
      "updated_at": "2021-06-12T12:31:58Z",
      "browser_download_url": "https://github.com/BurntSushi/ripgrep/releases/download/13.0.0/ripgrep-13.0.0-arm-unknown-linux-gnueabihf.tar.gz"
    },
    {
      "url": "https://api.github.com/repos/BurntSushi/ripgrep/releases/assets/38486879",
      "id": 38486879,
      "node_id": "MDEyOlJlbGVhc2VBc3NldDM4NDg2ODc5",
      "name": "ripgrep-13.0.0-i686-pc-windows-msvc.zip",
      "label": "",
    ...
      "content_type": "application/octet-stream",
      "state": "uploaded",
      "size": 1591463,
      "download_count": 3282,
      "created_at": "2021-06-12T12:32:47Z",
      "updated_at": "2021-06-12T12:32:48Z",
      "browser_download_url": "https://github.com/BurntSushi/ripgrep/releases/download/13.0.0/ripgrep-13.0.0-i686-pc-windows-msvc.zip"
    },
    {
      "url": "https://api.github.com/repos/BurntSushi/ripgrep/releases/assets/38486907",
      "id": 38486907,
      "node_id": "MDEyOlJlbGVhc2VBc3NldDM4NDg2OTA3",
      "name": "ripgrep-13.0.0-x86_64-apple-darwin.tar.gz",
      "label": "",
   ...
      "content_type": "application/octet-stream",
      "state": "uploaded",
      "size": 1815615,
      "download_count": 192942,
      "created_at": "2021-06-12T12:35:00Z",
      "updated_at": "2021-06-12T12:35:00Z",
      "browser_download_url": "https://github.com/BurntSushi/ripgrep/releases/download/13.0.0/ripgrep-13.0.0-x86_64-apple-darwin.tar.gz"
    },
    {
      "url": "https://api.github.com/repos/BurntSushi/ripgrep/releases/assets/38486889",
      "id": 38486889,
      "node_id": "MDEyOlJlbGVhc2VBc3NldDM4NDg2ODg5",
      "name": "ripgrep-13.0.0-x86_64-pc-windows-gnu.zip",
      "label": "",
     ...
      "content_type": "application/octet-stream",
      "state": "uploaded",
      "size": 9811405,
      "download_count": 10684,
      "created_at": "2021-06-12T12:33:25Z",
      "updated_at": "2021-06-12T12:33:26Z",
      "browser_download_url": "https://github.com/BurntSushi/ripgrep/releases/download/13.0.0/ripgrep-13.0.0-x86_64-pc-windows-gnu.zip"
    },
    {
      "url": "https://api.github.com/repos/BurntSushi/ripgrep/releases/assets/38486875",
      "id": 38486875,
      "node_id": "MDEyOlJlbGVhc2VBc3NldDM4NDg2ODc1",
      "name": "ripgrep-13.0.0-x86_64-pc-windows-msvc.zip",
      "label": "",
     ...
      "content_type": "application/octet-stream",
      "state": "uploaded",
      "size": 1734357,
      "download_count": 30550,
      "created_at": "2021-06-12T12:32:30Z",
      "updated_at": "2021-06-12T12:32:30Z",
      "browser_download_url": "https://github.com/BurntSushi/ripgrep/releases/download/13.0.0/ripgrep-13.0.0-x86_64-pc-windows-msvc.zip"
    },
    {
      "url": "https://api.github.com/repos/BurntSushi/ripgrep/releases/assets/38486871",
      "id": 38486871,
      "node_id": "MDEyOlJlbGVhc2VBc3NldDM4NDg2ODcx",
      "name": "ripgrep-13.0.0-x86_64-unknown-linux-musl.tar.gz",
      "label": "",
    ...
      "content_type": "application/octet-stream",
      "state": "uploaded",
      "size": 2109801,
      "download_count": 272291,
      "created_at": "2021-06-12T12:32:02Z",
      "updated_at": "2021-06-12T12:32:03Z",
      "browser_download_url": "https://github.com/BurntSushi/ripgrep/releases/download/13.0.0/ripgrep-13.0.0-x86_64-unknown-linux-musl.tar.gz"
    },
    {
      "url": "https://api.github.com/repos/BurntSushi/ripgrep/releases/assets/38493219",
      "id": 38493219,
      "node_id": "MDEyOlJlbGVhc2VBc3NldDM4NDkzMjE5",
      "name": "ripgrep_13.0.0_amd64.deb",
      "label": null,
     ...
      "content_type": "application/vnd.debian.binary-package",
      "state": "uploaded",
      "size": 1574096,
      "download_count": 71172,
      "created_at": "2021-06-12T17:36:20Z",
      "updated_at": "2021-06-12T17:36:21Z",
      "browser_download_url": "https://github.com/BurntSushi/ripgrep/releases/download/13.0.0/ripgrep_13.0.0_amd64.deb"
    }
  ],
```

mmohammadi9812: > Can we not let it error and give some type of output to the user to add an exe_name to the config?

I don't have many examples in my mind where the tool name is different than repo name and needs explicit user handling, so I don't think that benefits the end user.

The only case where I think manual handling would be beneficial, might be where two or more tools have the same name, or at least same repo name, which I think is a very rare case.

> Here is a small piece of the json returned by the repos/{owner}/repo/release/tags/{tag} endpoint.
> 
> The name is always formatted as `{name}-{tag}-{arch}.some.zip.extension`

Most of repos, especially the big ones, have a nice consistent format, but it's not forced and so, not always done consistently

An example of such would be [teeldear](https://github.com/dbrgn/tealdeer) where assets' names are in `{name}-{arch}.*` format:
```bash
:;curl -sL 'https://api.github.com/repos/dbrgn/tealdeer/releases' | jq '.[] | .assets | .[] | .name' | head -15
"completions_bash"
"completions_fish"
"completions_zsh"
"LICENSE-APACHE.txt"
"LICENSE-MIT.txt"
"tealdeer-linux-arm-musleabi"
"tealdeer-linux-arm-musleabi.sha256"
"tealdeer-linux-arm-musleabihf"
"tealdeer-linux-arm-musleabihf.sha256"
"tealdeer-linux-armv7-musleabihf"
"tealdeer-linux-armv7-musleabihf.sha256"
"tealdeer-linux-i686-musl"
"tealdeer-linux-i686-musl.sha256"
"tealdeer-linux-x86_64-musl"
"tealdeer-linux-x86_64-musl.sha256"
```

chshersh: The original issue #55 was about the ability to not specify the `exe_name` field when the `repo` field is specified. `repo` is a required field (unless the tool is hardcoded in `tool-sync`, in that case, you don't need to specify anything) so we always know the `repo` name. This way, we don't even need to query the GitHub API to get the `name`. We can just use `repo` from the config.

So I propose to move this discussion and the implementation away from the GitHub API-based implementation toward a simpler design 🙂

src/sync/configure.rs
```diff
@@ -31,7 +31,10 @@ pub fn configure_tool(tool_name: &str, config_asset: &ConfigAsset) -> Tool {
...
+    let exe_name = config_asset
+        .exe_name
+        .clone()
+        .unwrap_or(config_asset.repo.clone()?);
```
chshersh, line 37: `config_asset.repo.clone()?` is already evaluated to `repo` variable (that has type `String`) from the previous line so this can be simplified:
```suggestion
        .unwrap_or(repo);
```

src/sync/configure.rs
```diff
@@ -66,7 +69,7 @@ impl ToolInfo {
             exe_name: config_asset
                 .exe_name
                 .clone()
-                .unwrap_or_else(|| self.exe_name.clone()),
+                .unwrap_or(config_asset.repo.clone().unwrap()),
```
chshersh, line 72: This shouldn't be `config_asset.repo` unconditionally. To be more precise, the logic here should be:

1. If `config_asset.exe_name` is not empty, use it.
2. Otherwise (if it's empty), use `self.exe_name` if it's not empty.
3. If `self.exe_name` is empty, only then take the value of `repo`
    > The last step requires to move the value of the `repo` field inside a separate `let` definition before using the `ToolInfo` constructor so it can be reused in two places.

chshersh (changes requested):
Nice, that's exactly what I was looking for 👏🏻 

I just have a few suggestions to improve the implementation ✍🏻

src/sync/configure.rs
MitchellBerend: Oh I didnt know the release binary names aren't generated.

chshersh:
Hey @mmohammadi9812, do you need more help with this PR? I believe it's already done, just a few small things to be change 🙁 

Let me know if you have any questions!

chshersh:
@mmohammadi9812 I'm closing this PR in favour of:

* #104

I've taken your last commit from this PR and added my changes on top of it. But your contribution still counted! Thanks for taking your time to work on this issue 🤗

Issue #53, opened 2022-08-30
Consider using `human-panic` for better error reporting


popzxc:
Heya! Looking at [this file](https://github.com/chshersh/tool-sync/blob/main/src/err.rs#L16) it seems that you've imlpemented a small toolset for graceful exits upon meeting an irrecoverable error.

However, there exists a crate named [human-panic](https://crates.io/crates/human-panic) which serves the same functionality, but handles everything automatically and provides more info for debug. As a bonus, it'll also catch any other panic if it'll ever occur.

chshersh:
Thanks for the suggestion! I haven't heard about the `human-panic` package, it looks cool 👍🏻 
I guess, the only argument for not using it is to have fewer dependencies 🤔 

Also, in my file I'm using `process::exit(1)` instead of `panic!("...")`. Will there be any difference?

popzxc:
Well, Rust uses error code 101 for panics. I don't think there should be any other differences related to the process termination.

Issue #25, opened 2022-08-24
Provide more detailed TOML decoding errors


chshersh:
Currently, when TOML decoding fails due to missing fields or mismatched values, the error doesn't specify the missing field:

```
Error parsing configuration at path Cargo.toml: [Decode Error]
```

The corresponding function also simply returns `Option<Config>` instead of providing more details

https://github.com/chshersh/tool-sync/blob/950b4028a619d33303548acd06056949dcec1a4a/src/config/toml.rs#L48-L64

It would provide better UX to add more details to the `DecodeError` constructor in the `TomlError` type by telling all errors that happened during decoding:

https://github.com/chshersh/tool-sync/blob/950b4028a619d33303548acd06056949dcec1a4a/src/config/toml.rs#L10-L14

Open questions:

* Does the `toml` crate provide better decoding than `as_str()` and friends? All such functions return `Option<...>` while it would be better to have `Result<..., ...>`
* Does Rust have the analogue of the Haskell's `Validation` type to combine multiple errors and which can be used with `toml`?

DukeManh:
Hey @chshersh, I ran into the same issue trying your tool. I'd like help with this.

chshersh:
@DukeManh Sure, go for it!

PR #78, opened 2022-09-10
[#73] Adds build.rs to generate a default config template with tool version

### LINKED ISSUE #73: Add current version to the generated config without hard coding it in the actual template.
Referencing #62 
> Would be cool to also say by which version of tool-sync but not sure how to do this without hardcoding the version 

> We can do some magic in a build.rs file that can generate the src/config/template.rs so it contains the version in the config template itself.

Im not quite sure if a `build.rs` is the best way to implement this since this means we have to have a template for the `src/config/template.rs` file itself. This sounds like a very weird approach to me but I currently don't see any other way to do this.

The `build.rs` file would look something like
```rust

const TEMPLATE_RS_TEMPLATE: &str = r##"
The actual template
"##


fn main() -> std::io::Result<()> {
  std::fs::write_all("src/config/template.r", TEMPLATE_RS_TEMPLATE)?;
}

```

MitchellBerend:
Resolves #73
This adds a build script that generates the `src/config/template.rs` file. It is
done this way because there is no access to version information after compile
time. It uses the `CARGO_PKG_VERSION` environment variable made available for
cargo after running the command `cargo build`.

### Additional tasks

tasks following #73

---

- [x] Automatically embedding version. As discussed in this issue.
- [x] Prefill all the tools supported by tool-sync. This requires converting the
    big match to a BTreeMap.
- [x] Add comments around tool tables. To make sure that the code inside the
    comments stays up-to-date and is parsed by tool-sync itself.

General tasks

---

- [x] Tests added


build.rs
```diff
@@ -0,0 +1,62 @@
...
+#
+# uncomment if you want to install on macOS as well
+# asset_name.macos = "apple-darwin"
+#
+# uncomment if you want to install on Windows as well
+# asset_name.windows = "x86_64-pc-windows-msvc"
+"##;
+
+fn main() -> Result<()> {
+    let s = format_template();
```
MitchellBerend, line 43: I need to change this variable name still actually.

chshersh (changes requested):
Thanks for attempting to do this!

It's an interesting approach 🤔 But I tend to think that we don't need compile-time generation in this particular case.

I imagine using the [env!](https://doc.rust-lang.org/std/macro.env.html) macro to read the value of the `CARGO_PKG_VERSION` env variable and `format!` inside the `src/config/template.rs` to produce a final default configuration.

It'll become uglier than having just a plain hardcoded string but I can live with a bit of ugliness 👍🏻  The existing approach won't work if we want to fill all the hardcoded tools from a `BTreeMap`. To do this with `build.rs`, we had to move this map here and the implementation will become more convoluted.

I do, however, see the benefit in having the full default config written as plain text. So I propose to use golden testing here. Let's create the `tests/default.toml` file with the hardcoded config as it will be generated. And add a test that reads the file and compares it with the result of generating the config.

This is known as the golden testing technique. It's nice to have the full text to visibly see the content. The important part here is that we also test this text against the result of generation to make sure that we do generate the valid config.

src/sync/db.rs
```diff
@@ -1,10 +1,23 @@
+use std::collections::BTreeMap;
+
 use crate::model::asset_name::AssetName;
 use crate::model::tool::{ToolInfo, ToolInfoTag};
 
 /// Get info about known tools from a hardcoded database
 pub fn lookup_tool(tool_name: &str) -> Option<ToolInfo> {
-    match tool_name {
-        "bat" => Some(ToolInfo {
+    let mut known_db = build_db();
```
chshersh, line 8: If I'm not mistaken, this will rebuild the entire `BTreeMap` from scratch on each tool. This is probably not noticeable but it's very easy to avoid this. Instead of having the `build_db()` function, we can create a top level `const db: BTReeMap...` using the `from` method:

* https://doc.rust-lang.org/std/collections/struct.BTreeMap.html#impl-From%3C%5B(K%2C%20V)%3B%20N%5D%3E

src/config/template.rs
```diff
@@ -1,12 +1,16 @@
...
+        "# This config file was generated for version {version}
+#
+# # tool-sync default configuration file
 # https://github.com/chshersh/tool-sync
 # This file was automatically generated by tool-sync
 #####################################################
 #
 #
-# store_directory = "$HOME/.local/bin"
+# store_directory = \"$HOME/.local/bin\"
```
chshersh, line 13: Why not use raw string literals `r##" ...  "##` and avoid escaping quotes?

chshersh (changes requested):
Looks great! 👍🏻 

Just a few comments 🙂

src/config/template.rs
MitchellBerend: I could have sworn I tried this and it did some funny business with the `{version}`. Trying this now it works.

src/sync/db.rs
MitchellBerend: > If I'm not mistaken, this will rebuild the entire BTreeMap from scratch on each tool.

correct, there is no const method to create a BTreeMap though. [Rust playground example](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=ee1dc50e411e177b9902083b5d981684).

MitchellBerend: There is actually a crate called [lazy_static](https://crates.io/crates/lazy_static) that provides a macro to do something like this but im not sure if this actually creates a const.
It would also add another dependency.

MitchellBerend:
I must have clicked a ready button or something while reading some comments on this pr. It still needs a lot of work and I would also like to add a doc tests for the output of the `default-config` command

src/sync/db.rs
chshersh: Aha, I see. Learned something new today 😅 
Didn't know you need to declare a function as `const` for it to be used in `const` variable 🧐 

In that case, the existing approach is fine 👍🏻 Not sure if I want to bring an extra dependency if it doesn't really solve anything. We can deal with it when this map becomes huge and starts bringing some performance problems but I don't expect this in the near future 😌

MitchellBerend: Ye this map should be fine until it grows by orders of magnitude I think.

MitchellBerend:
Im adding some more tests to increase coverage. Im using [cargo tarpaulin](https://github.com/xd009642/tarpaulin) to check this. 
There are still some false negatives in tarpaulin but adding more tests is probably not a bad thing.

I have found that this method was not testable since it exits the program:
https://github.com/chshersh/tool-sync/blob/4341cc734142fda35caacdd5c02d778d00698144/src/config/schema.rs#L45-L63


main brach
---
```bash
Sep 11 17:10:36.148  INFO cargo_tarpaulin::process_handling::linux: Launching test
Sep 11 17:10:36.148  INFO cargo_tarpaulin::process_handling: running /home/mitchell/rust/tool-sync/target/debug/deps/tool-711d73e056f99e07

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

Sep 11 17:10:38.071  INFO cargo_tarpaulin::process_handling::linux: Launching test
Sep 11 17:10:38.071  INFO cargo_tarpaulin::process_handling: running /home/mitchell/rust/tool-sync/target/debug/deps/tool_sync-fd5c4dc2ba40d47f

running 21 tests
test config::toml::tests::empty_file ... ok
test config::toml::tests::store_directory_is_a_number ... ok
test config::toml::tests::only_store_directory ... ok
test config::toml::tests::single_empty_tool ... ok
test config::toml::tests::single_partial_tool ... ok
test config::toml::tests::single_full_tool ... ok
test config::toml::tests::store_directory_is_dotted ... ok
test config::toml::tests::two_empty_tools ... ok
test model::asset_name::tests::exe_name ... ok
test model::asset_name::tests::asset_name ... ok
test sync::configure::tests::full_override ... ok
test sync::configure::tests::full_configuration ... ok
test sync::configure::tests::known_tool_with_empty_config_asset ... ok
test sync::configure::tests::unknown_tool_with_empty_config_asset ... ok
test sync::configure::tests::partial_configuration ... ok
test sync::configure::tests::wrong_tool_with_empty_config_asset ... ok
test sync::download::tests::release_url_with_latest_tag_is_correct ... ok
test sync::configure::tests::partial_override ... ok
test sync::download::tests::release_url_with_specific_tag_is_correct ... ok
test sync::progress::tests::test_max_tag_size_specific ... ok
test sync::progress::tests::test_max_tag_size_latest ... ok

test result: ok. 21 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.05s

Sep 11 17:10:40.028  INFO cargo_tarpaulin::report: Coverage Results:
|| Uncovered Lines:
|| src/config/schema.rs: 45-46, 48-50, 53, 55-56, 58
|| src/config/toml.rs: 17-21, 26-27, 29
|| src/err.rs: 4-5, 12, 16-17, 30
|| src/lib.rs: 16-18, 20-26, 29-30, 33, 37-45, 47-48, 54-55
|| src/main.rs: 2
|| src/model/tool.rs: 19-22, 24
|| src/sync/archive.rs: 28-33, 40, 46, 48-53, 55-56, 58-63, 65-66, 68-72, 81-82, 84, 87-89, 93-95, 101, 103-106, 109-110, 112, 114, 117, 123, 126, 128-130, 133-134, 138, 140, 145-146, 148-151
|| src/sync/configure.rs: 22
|| src/sync/db.rs: 8-14, 16, 19-25, 27, 30-36, 38, 41-47, 49, 63-69, 71
|| src/sync/download.rs: 31-33, 37-38, 40-42, 46-47, 50-52, 55, 57, 60-61, 64-66, 69, 71-72, 74-75, 77-80, 83-84, 87-88, 90, 94-95, 97, 99-100, 102, 104-107, 109-112, 119-122
|| src/sync/install.rs: 29-33, 43-45, 47-50, 52-54, 57-59, 64, 69, 71, 73, 75-78, 83, 87-89, 92, 94, 96-100, 109-110, 112-114, 117, 119, 121, 128-129
|| src/sync/progress.rs: 38-39, 47, 50-51, 53-56, 60-61, 64-65, 68-69, 72-73, 75-77, 80-81, 83-85
|| src/sync.rs: 12-14, 32, 34-35, 38, 40-41, 43-44
|| Tested/Total Lines:
|| src/config/schema.rs: 0/9 +0.00%
|| src/config/toml.rs: 71/79 -10.13%
|| src/err.rs: 0/6 +0.00%
|| src/lib.rs: 0/26 +0.00%
|| src/main.rs: 1/2 +0.00%
|| src/model/asset_name.rs: 16/16 +0.00%
|| src/model/tool.rs: 4/9 +0.00%
|| src/sync/archive.rs: 0/63 +0.00%
|| src/sync/configure.rs: 108/109 +0.00%
|| src/sync/db.rs: 16/56 -71.43%
|| src/sync/download.rs: 14/66 +0.00%
|| src/sync/install.rs: 0/47 +0.00%
|| src/sync/progress.rs: 29/54 +0.00%
|| src/sync.rs: 0/11 +0.00%
|| 
46.84% coverage, 259/553 lines covered, -12.641681145031853% change in coverage
```



This branch
---
```bash
Sep 11 17:15:10.011  INFO cargo_tarpaulin::process_handling::linux: Launching test
Sep 11 17:15:10.011  INFO cargo_tarpaulin::process_handling: running /home/mitchell/rust/tool-sync/target/debug/deps/tool-711d73e056f99e07

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

Sep 11 17:15:11.899  INFO cargo_tarpaulin::process_handling::linux: Launching test
Sep 11 17:15:11.899  INFO cargo_tarpaulin::process_handling: running /home/mitchell/rust/tool-sync/target/debug/deps/tool_sync-fd5c4dc2ba40d47f

running 27 tests
test config::toml::tests::store_directory_is_a_number ... ok
test config::toml::tests::empty_file ... ok
test config::toml::tests::only_store_directory ... ok
test config::toml::tests::single_empty_tool ... ok
test config::toml::tests::single_partial_tool ... ok
test config::toml::tests::single_full_tool ... ok
test config::toml::tests::test_toml_error_display_decode ... ok
test config::toml::tests::test_toml_error_display_io ... ok
test config::toml::tests::test_parse_file_error ... ok
test config::toml::tests::test_parse_file_correct_output ... ok
test config::toml::tests::store_directory_is_dotted ... ok
test config::toml::tests::test_parse_file_passing_tools_read ... ok
test config::toml::tests::two_empty_tools ... ok
test model::asset_name::tests::exe_name ... ok
test config::toml::tests::test_toml_error_display_parse ... ok
test model::asset_name::tests::asset_name ... ok
test sync::configure::tests::unknown_tool_with_empty_config_asset ... ok
test sync::configure::tests::partial_configuration ... ok
test sync::configure::tests::known_tool_with_empty_config_asset ... ok
test sync::configure::tests::partial_override ... ok
test sync::configure::tests::full_override ... ok
test sync::configure::tests::full_configuration ... ok
test sync::configure::tests::wrong_tool_with_empty_config_asset ... ok
test sync::progress::tests::test_max_tag_size_specific ... ok
test sync::progress::tests::test_max_tag_size_latest ... ok
test sync::download::tests::release_url_with_specific_tag_is_correct ... ok
test sync::download::tests::release_url_with_latest_tag_is_correct ... ok

test result: ok. 27 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.05s

Sep 11 17:15:13.851  INFO cargo_tarpaulin::report: Coverage Results:
|| Uncovered Lines:
|| src/config/schema.rs: 45-46, 48-50, 53, 55-56, 58
|| src/config/template.rs: 3-4
|| src/err.rs: 4-5, 12, 16-17, 30
|| src/lib.rs: 16-18, 20-26, 29-30, 33, 37-45, 47-48, 54-55
|| src/main.rs: 2
|| src/model/tool.rs: 19-22, 24
|| src/sync/archive.rs: 28-33, 40, 46, 48-53, 55-56, 58-63, 65-66, 68-72, 81-82, 84, 87-89, 93-95, 101, 103-106, 109-110, 112, 114, 117, 123, 126, 128-130, 133-134, 138, 140, 145-146, 148-151
|| src/sync/configure.rs: 22
|| src/sync/download.rs: 31-33, 37-38, 40-42, 46-47, 50-52, 55, 57, 60-61, 64-66, 69, 71-72, 74-75, 77-80, 83-84, 87-88, 90, 94-95, 97, 99-100, 102, 104-107, 109-112, 119-122
|| src/sync/install.rs: 29-33, 43-45, 47-50, 52-54, 57-59, 64, 69, 71, 73, 75-78, 83, 87-89, 92, 94, 96-100, 109-110, 112-114, 117, 119, 121, 128-129
|| src/sync/progress.rs: 38-39, 47, 50-51, 53-56, 60-61, 64-65, 68-69, 72-73, 75-77, 80-81, 83-85
|| src/sync.rs: 12-14, 32, 34-35, 38, 40-41, 43-44
|| Tested/Total Lines:
|| src/config/schema.rs: 0/9 +0.00%
|| src/config/template.rs: 0/2
|| src/config/toml.rs: 117/117 +10.13%
|| src/err.rs: 0/6 +0.00%
|| src/lib.rs: 0/26 +0.00%
|| src/main.rs: 1/2 +0.00%
|| src/model/asset_name.rs: 16/16 +0.00%
|| src/model/tool.rs: 4/9 +0.00%
|| src/sync/archive.rs: 0/63 +0.00%
|| src/sync/configure.rs: 108/109 +0.00%
|| src/sync/db.rs: 73/73 +71.43%
|| src/sync/download.rs: 14/66 +0.00%
|| src/sync/install.rs: 0/47 +0.00%
|| src/sync/progress.rs: 29/54 +0.00%
|| src/sync.rs: 0/11 +0.00%
|| 
59.34% coverage, 362/610 lines covered, +12.508819257107277% change in coverage
```

chshersh:
@MitchellBerend `cargo tarpaulin` looks like a nice tool 🙂 I wouldn't try to gamify code coverage too much. Achieving 100% actually might be counterproductive as your test freeze your code too much and you need to do more and more to change the tests each time you change the code. At this stage, `tool-sync` is quite _unstable_ and lots of refactorings could happen by changing the codebase significantly.

It's useful however to identify important code paths that are not covered 👍🏻 

> Should the default tools also be generated in the default-config command? This would make it so maintenance only needs to happen in one place (the BTreeMap).

Yes, that was the plan. If you're up to it, you can go ahead and do this change in this PR (or leave it for later).

src/config/template.rs
```diff
@@ -1,6 +1,22 @@
...
-pub const CONFIG_TEMPLATE: &str = r##"# # tool-sync default configuration file
+use crate::sync::db::build_db;
+
+pub fn config_template() -> String {
+    let mut tools: String = String::new();
+    for tool in build_db().keys().cloned().collect::<Vec<String>>() {
+        if let Err(e) = writeln!(tools, "# [{}]", tool) {
+            panic!("{}", e);
+        };
+    }
```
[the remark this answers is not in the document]
chshersh: I guess I used to think more in a functional style, so I would write this code like this, trying to be closed to idiomatic Rust as well:

```rust
    let tools = build_db()
        .keys()
        .cloned()
        .fold(String::new(), |mut acc, k| {
            acc.reserve(k.len() + 5);
            acc.push_str("# [");
            acc.push_str(&k);
            acc.push_str("]\n");
            acc
        });
```

But the current implementation is fine as well if we want to squeeze the most performance 👍🏻 

My only suggestion would be to use `abort_suggest_issue` instead of `panic!`.

<hr>

There's also an open issue about using `human-panic`:

* #53 

But I haven't looked into this library yet so I don't have opinions about `human-panic` despite being an extra dependency.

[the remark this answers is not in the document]
MitchellBerend: > But I haven't looked into this library yet so I don't have opinions about human-panic despite being an extra dependency.

This is the first time I have heard of this lib actually so no input from me on this.

MitchellBerend:
I think this is ready for review. I would like to have more tests still though but I agree with the not having too high of a code coverage. The tests I added were to confirm the formatting of printing errors when processing toml files so I would like those in.

.github/workflows/ci.yml
```diff
@@ -89,6 +89,17 @@ jobs:
...
+    - if: matrix.os != 'windows-latest'
+      name: "Characterization test: [unix] [default-config]"
+      env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      run: |
+          cargo run -- default-config > tools.toml
+
+          diff tools.toml tests/default-config.toml
+
+          if [[ $? == 1 ]]; then cat tools.toml; false; fi
```
chshersh, line 101: What if we just keep `diff` as the last command? I feel like having the `diff` output in the end might be more helpful in seeing a specific error from CI logs

src/config/template.rs
```diff
@@ -1,6 +1,22 @@
...
+
+pub fn config_template() -> String {
+    let mut tools: String = String::new();
+    for tool in build_db().keys().cloned().collect::<Vec<String>>() {
+        if let Err(e) = writeln!(tools, "# [{}]", tool) {
+            crate::err::abort_suggest_issue(&format!("{}", e));
+        };
+    }
+    // adding another hash to fil a new line before the next block
+    tools.push('#');
```
chshersh, line 14: We can just add `#` in the template string directly and don't push it to a string here. We already know that `build_db()` returns a non-empty Map so we don't need to handle this special case when `tools` might be empty

src/sync/db.rs
```diff
@@ -1,10 +1,23 @@
...
+    match known_db.get_mut(tool_name) {
+        Some(tool_info) => tool_info.take(),
+        None => None,
+    }
+}
+
+pub fn build_db() -> BTreeMap<String, Option<ToolInfo>> {
```
chshersh, line 15: I wonder why it's a `BTreeMap` to `Option<ToolInfo>`? I see that all keys use the `Some` constructor. Do you see a case where we would map a tool name to `None`?

src/config/toml.rs
```diff
@@ -103,6 +103,89 @@ fn str_by_key(table: &Map<String, Value>, key: &str) -> Option<String> {
...
     use super::*;
 
+    #[test]
+    fn test_toml_error_display_io() {
+        let toml_error = TomlError::IO(String::from("some file error!"));
+
+        assert_eq!(
+            String::from("[IO Error] some file error!"),
+            toml_error.display()
+        );
```
chshersh, line 113: Generally, testing the exact rendered error messages as strings is considered an anti-pattern. Because if you change the string definition, you now need to do a busy work of updating tests but those tests don't actually verify anything. Textual error messages is a piece of display information and it's already written in the code so repeating it doesn't bring much to the table.

It makes sense, however, to define errors as enumerations and test that specific preconditions lead to the expected error constructors.

For example, this particular test would be more helpful if the function tried to read a non-existing file and we `assert` that it returns `TomlError::IO` without checking the actual content of this error (because the message is not particularly interesting).

src/config/toml.rs
```diff
@@ -103,6 +103,89 @@ fn str_by_key(table: &Map<String, Value>, key: &str) -> Option<String> {
...
+                assert_eq!(
+                    String::from(
+                        "[Parsing Error] expected an equals, found an identifier at line 1 column 8"
+                    ),
+                    toml_error.display()
+                );
+            }
+            Ok(_) => unreachable!(),
+        };
+    }
```
chshersh, line 137: Same as with `TomlError::IO` and I also think there's an easier way to test this error. You can pass `"broken toml"` directly to the `parse_string` function

src/config/toml.rs
```diff
@@ -103,6 +103,89 @@ fn str_by_key(table: &Map<String, Value>, key: &str) -> Option<String> {
...
+
+    #[test]
+    fn test_toml_error_display_decode() {
+        let toml_error = TomlError::Decode;
+        assert_eq!(String::from("[Decode Error]"), toml_error.display());
+    }
```
chshersh, line 143: Unfortunately, not so helpful error 😞 
There's an issue though to provide more detailed errors:

* https://github.com/chshersh/tool-sync/issues/25

src/config/toml.rs
```diff
@@ -103,6 +103,89 @@ fn str_by_key(table: &Map<String, Value>, key: &str) -> Option<String> {
...
+
+    #[test]
+    fn test_parse_file_correct_output() {
+        let test_config_path = PathBuf::from("tests/full-database.toml");
+        let config = parse_file(&test_config_path).expect("This should not fail");
+
+        assert_eq!(String::from("full-database"), config.store_directory);
+    }
```
chshersh, line 151: This is actually a good test because it actually tests parsing of a config from a file path and currently we don't have a test for this 👍🏻 

In my previous tests I've hardcoded TOML content

src/config/toml.rs
```diff
@@ -103,6 +103,89 @@ fn str_by_key(table: &Map<String, Value>, key: &str) -> Option<String> {
...
+        match parse_file(&test_config_path) {
+            Ok(_) => {
+                unreachable!()
+            }
+            Err(_) => {
+                final_assert = true;
+            }
+        };
+        assert!(final_assert);
+    }
```
chshersh, line 168: I'm a bit surprised to learn there's nothing like `assert_if` in a standard library that can take a predicate and match on a constructor 🤔 

The current way of writing this test feels a bit awkward to me but it's hard to produce a better version 🥴 

What do you think about?

```rust
        match parse_file(&test_config_path) {
            Ok(_) => {
                assert!(false, "Unexpected success")
            }
            Err(_) => {
                assert!(true, "Expected a parsing error")
            }
        }

File: src/config/toml.rs [hunk 13]
```diff
@@ -103,6 +103,89 @@ fn str_by_key(table: &Map<String, Value>, key: &str) -> Option<String> {
...
+            "bat".into(),
+            "difftastic".into(),
+            "exa".into(),
+            "fd".into(),
+            "ripgrep".into(),
+            "tool-sync".into(),
+        ];
+
+        assert_eq!(correct_vec, tools_vec);
+    }
```
chshersh, line 187: In this particular shape, this test is not useful. I would move it into the `db` module and check that the result of parsing `full_database` is the same as `build_db().keys().cloned().collect()` so we can make sure that this test case actually covers all hard-coded tools

chshersh (changes requested):
Great work, as always! 🥇 

It's nice to see the conversion to `BTreeMap` of all the tools, it opens new possibilities 🙂 

I have a few comments and some clarification questions but overall it looks really great!

src/config/template.rs
MitchellBerend: `writeln` also appends a `\n` to the end of the string so removing this hash would mean the config looks like this:

```bash
#
# [bat]
# [difftastic]
# [exa]
# [fd]
# [ripgrep]
# [tool-sync]

# To add configuration for other tools these are the config options:
# [ripgrep]
#        owner     = "BurntSushi"
#        repo      = "ripgrep"
#        exe_name  = "rg"

```

src/sync/db.rs
MitchellBerend: This is an oversight on my part, I just copy pasted the some variant from the match statement. You are right that there is no reason for this.

src/config/toml.rs
MitchellBerend: I like this idea, having a proper error message when tests fail is way better than hitting some `unreachable!()` block and having to look at the code to figure out what happened.

src/config/toml.rs
MitchellBerend: My intention was actually to test for the string within the brackets. I can change this one to only check for the actual error, or another option is adding the check along side the string check.

src/config/toml.rs
MitchellBerend: Same thing as above. The passing the string to the function itself is a better idea though.

src/config/toml.rs
MitchellBerend: Do you have a suggestion to avoid the `expect()`?

I think this will only fail if there is some io error in ci, right?

src/config/toml.rs
MitchellBerend: Im removing this test entirely since we have the test that compares the output of the `default-config` command with `tests/default-config.toml`. This would actually be the same as checking if all tools are present.

src/config/toml.rs
chshersh: I think usages of `unwrap()` and `expect()` (expect is better) are fine in tests.

If they fail, the whole test will fail as well (I assume) which means that the test is incorrect. Ideally we shouldn't have fleaky tests in the codebase but this one looks pretty deterministic to me so it's okay 👌🏻

src/config/template.rs
chshersh: Aha, I see now 👀

src/sync/db.rs
chshersh: Great! I'm glad I understand this correctly 😅 

I was also thinking if `lookup_tool` could be just defined as `build_db().get(tool_name)` but apparently there's some reference business preventing this from happening 🤔

MitchellBerend: There is actually a small reference problem that can be circumvented by removing the key from the `HasmMap` with `remove(key)`. That means tools can't be downloaded twice in the same `lookup_tool` function call.

I don't see this as a problem but its good to have this known ahead of time before someone spends hours of their life looking for this particular issue.

src/config/toml.rs
MitchellBerend: I wrapped this `expect()` call in a `std::panic::catch_unwind` so if there is a panic in the test will not fail. Im not sure if this is the right approach though since the result of this test might not be accurate if it does.

chshersh: I was okay with the previous usage of `expect` as well 👍🏻 
Sorry if I was vague in my comment 😅 

I think, there's a better way to write this test without using panics at all. Something like:

```rust
let test_config_path = PathBuf::from("tests/full-database.toml");

match parse_file(&test_config_path) {
    Err(_) => assert!(false, "Unexpected error parsing: tests/full-database.toml"),
    Ok(config) => assert_eq!(String::from("full-database"), config.store_directory),
}
```

MitchellBerend: > Sorry if I was vague in my comment sweat_smile
It wasn't, I would rather this discussion happen anyway.

Does this test not still potentially fail without actually testing what needs to be tested? Or is this a case of, it doesn't matter how its implemented? It will either be a false positive or a false negative.

.github/workflows/ci.yml
```diff
@@ -89,6 +89,17 @@ jobs:
...
+    - if: matrix.os != 'windows-latest'
+      name: "Characterization test: [unix] [default-config]"
+      env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      run: |
+          cargo run -- default-config > tools.toml
+
+          diff tools.toml tests/default-config.toml
+
+          if [[ $? == 1 ]]; then tools.toml tests/default-config.toml; false; fi
```
chshersh, line 101: Should we call `diff` here?
```suggestion
          if [[ $? == 1 ]]; then diff tools.toml tests/default-config.toml; false; fi
```

I also wonder if we can completely remove this line. `diff` will fail with error code 1 if there's a difference and print the diff. So we can just leave the call to `diff` in the end
```suggestion
```

MitchellBerend: Ye this is a typo

chshersh (approved):
I believe it's good to go 🚢 
Thanks a lot for your huge work, @MitchellBerend 🏆 

Nothing is immutable so the code always can be improved later in case we have more ideas 🙂 
But this PR is really helpful.

PR #80, opened 2022-09-11
[#33] Better end-of-sync output

### LINKED ISSUE #33: Better end-of-sync output
A few improvements ideas:

- [ ] Mention in the end how many tools downloaded and how many failed
- [ ] Suggest to change config/open an issue on specific errors for failed tools

chshersh:
Resolves #33

This PR changes the fetching algorithm significantly in an attempt to improve the final output. Specifically,

1. Information about assets of all tools is fetched first before downloading the tools. This way users see all validation errors before the downloading has started.
    > The only validation errors left are missing `exe` files in assets, I believe. Or GitHub Rate-Limit error when trying to download an asset.
2. Estimated download size is printed before the downloading (so users can see how much it would take to download).
3. The correct tags are shown immediately.
4. Lots of refactoring.

A single GIF worth a thousand words:

gif_1

### DIFF
--- a/src/model/release.rs
+++ b/src/model/release.rs
@@ -6,7 +6,7 @@ pub struct Release {
     pub assets: Vec<Asset>,
 }
 
-#[derive(Deserialize, Debug)]
+#[derive(Deserialize, Debug, Clone)]
 pub struct Asset {
     pub id: u32,
     pub name: String,
--- a/src/sync.rs
+++ b/src/sync.rs
@@ -3,16 +3,29 @@ mod configure;
 mod db;
 mod download;
 mod install;
+mod prefetch;
 mod progress;
 
+use console::Emoji;
+
 use crate::config::schema::Config;
-use crate::sync::install::Installer;
-use crate::sync::progress::SyncProgress;
+
+use self::install::Installer;
+use self::prefetch::prefetch;
+use self::progress::SyncProgress;
+use self::progress::ToolPair;
 
 pub fn sync(config: Config) {
     if config.tools.is_empty() {
-        eprintln!(
-            r#"No tools to sync. Have you configured 'tool-sync'?
+        no_tools_message()
+    } else {
+        sync_tools(config)
+    }
+}
+
+fn no_tools_message() {
+    eprintln!(
+        r#"No tools to sync. Have you configured 'tool-sync'?
 
 Put the following into the $HOME/.tool.toml file for the simplest configuration:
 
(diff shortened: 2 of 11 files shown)


src/model/release.rs
```diff
@@ -12,3 +12,13 @@ pub struct Asset {
...
+
+impl From<&Asset> for Asset {
+    fn from(other: &Asset) -> Self {
+        Asset {
+            id: other.id,
+            name: other.name.clone(),
+            size: other.size,
+        }
+    }
+}
```
chshersh, line 24: I don't know what is the best way to convert from `&A` to `A` so found usages of the `From` trait.

I was also thinking about [ToOwned](https://doc.rust-lang.org/stable/std/borrow/trait.ToOwned.html) but to me `ToOwned` looked like it should be used for different types (e.g. `&str / String`, `Path / PathBuf`)

src/sync/progress.rs
```diff
@@ -10,19 +10,26 @@ pub struct SyncProgress {
...
+pub struct ToolPair<'a> {
+    pub name: &'a str,
+    pub tag: &'a str,
+}
+
 impl SyncProgress {
     /// Creates new `SyncProgress` from a list of tools.
     /// !!! The given `Vec` must be non-empty !!!
-    pub fn new(tools: Vec<String>, tags: Vec<String>) -> SyncProgress {
+    pub fn new(tool_assets: Vec<ToolPair>) -> SyncProgress {
```
chshersh, line 25: Oops, this should be `tool_pairs` instead of `tool_assets`

src/sync.rs
```diff
@@ -27,21 +40,34 @@ Put the following into the $HOME/.tool.toml file for the simplest configuration:
...
+    let mut installed_tools: u64 = 0;
+
+    for tool_asset in tool_assets {
+        let is_successs = installer.install(tool_asset);
+        if is_successs {
+            installed_tools += 1
         }
     }
+
+    eprintln!("{} Successfully installed {} tools!", DONE, installed_tools);
```
chshersh, line 72: Probably can also say where exactly they were installed 🤔 
Sometimes, when locally testing with different configs, I don't remember all the directories 🥴

src/model/release.rs
MitchellBerend: How I understand `to_owned` is that it is essentially a superset or `clone`, I am for putting in the smallest scope so the clone way would be best.

src/sync/prefetch.rs
```diff
@@ -0,0 +1,157 @@
...
+
+    prefetch_progress.finish();
+
+    let estimated_download_size: u64 = tool_assets.iter().map(|ta| ta.asset.size).sum();
+    let size = HumanBytes(estimated_download_size);
+    eprintln!(
+        "{emoji} Estimated download size: {size}",
+        emoji = PACKAGE,
+        size = size
+    );
```
MitchellBerend, line 93: Maybe we should include that this is the size of the sum of all tools. This was not clear to me immediately so I had to look that up in the code.

MitchellBerend:
Im happy that the structure of the fetching and actual downloading is the way it is in this change. It means actual decisions can be made based on the status of tools in the configuration.

This might also enable a prediction about how many of the configured tools can actually be downloaded since there is a rate limit of 60 requests per hour. I think the hard limit we run into right now is 30 tools total. There is 1 api call to find the id of the tag and then 1 call to actually download it.

src/model/release.rs
chshersh: Thanks for the suggestion! Wasn't sure `Clone` could work here. It simplified code indeed 😌

chshersh:
> It means actual decisions can be made based on the status of tools in the configuration.

Indeed, that's the plan 🙂 

> This might also enable a prediction about how many of the configured tools can actually be downloaded since there is a rate limit of 60 requests per hour. I think the hard limit we run into right now is 30 tools total. There is 1 api call to find the id of the tag and then 1 call to actually download it.

If you use a GitHub token, the limit is actually 5000 (and not 60). I believe we have some capacity until people need more than 2500 tools 👍🏻

MitchellBerend:
Turns out my `GITHUB_TOKEN` had a space in it :sweat:.

MitchellBerend (approved):

chshersh:
Here is the final GIF after applying suggestions in this PR:

gif_1

Issue #82, opened 2022-09-14
does not work behind a proxy (windows)


maku:
I tried tool-sync on 2 machines. On the machine which uses a proxy the `tool sync` command does not work...

Is there a way to config tool-sync for using a proxy?

chshersh:
Hi @maku 👋🏻 

Currently `tool-sync` doesn't support proxy. I'm using the `ureq` Rust crate for querying an API. I see that it supports proxies:

* [Using HTTP CONNECT](https://github.com/algesten/ureq#example-using-http-connect)
* [`Proxy` struct](https://docs.rs/ureq/latest/ureq/struct.Proxy.html)

But I'm not quite sure how it should be used 🤔 Could you help me out here and provide some more details? What CLI and UX would you expect from `tool-sync`?

Or, alternatively, how would you change the following `curl` command to download an asset if you wanted to use proxies?

```shell
curl \
    -H "Accept: application/vnd.github+json"  \
    https://api.github.com/repos/chshersh/tool-sync/releases/assets/76226041
```

maku:
I would suggest to provide a configuration option for the proxy string in the .tool.toml file 

e.g. 
`proxy = http://127.0.0.1:8080`

and pass the data to ureq when available

MitchellBerend:
@maku are you working on this? Otherwise I will try and implement this. Im actually running into the same issue right now.

maku:
@MitchellBerend no i do not work on this

MitchellBerend:
I have tried playing around with with this issue and its actually harder than I thought.

Some thoughts I have already:

There should probably be a global proxy option] but I can also see different tools having different proxies.
A new flag `--proxy` seems like a nice way to adhoc add a proxy for a single run just like `--config`. 

What do you guys think? 

Example configs
```bash
# This configuration is automatically generated by tool-sync 0.2.0
# https://github.com/chshersh/tool-sync
#######################################
#
# Installation directory for all the tools:
store_directory = "$HOME/rust/tool-sync/bin"
proxy = "http://10.10.0.1:8000"
#
# tool-sync provides native support for some of the tools without the need to
# configure them. Uncomment all the tools you want to install with a single
...
```


```bash
[github]  # Name of the tool (new or one of the hardcoded to override default settings)
    owner     = "cli"  # GitHub repository owner
    repo      = "cli"     # GitHub repository name
    exe_name  = "gh"          # Executable name inside the asset
#
#     Uncomment to download a specific version or tag.
[...]

PR #89, opened 2022-09-15
[#57] Adds install command to install known tools

### LINKED ISSUE #57: Implement the 'tool install' command
This command can install a single tool from the repo and update the default configuration as well.

In the first version of this command, it should only support the installation of known tools with the command like:

```shell
$ tool install bat
```

If the tool is unknown, it should print the error message:

```shell
$ tool install xxx
Unknown tool: 'xxx'
Supported tools:
    * bat
    * ripgrep
    * ...
```

The specific behaviour:

1. Check if the tool is known.
2. If yes, try syncing the tool.

MitchellBerend:
Resolves #57 

This pr adds an `isntall` subcommand that lets the user install any of the hardcoded tools. As a bonus it also implements the `From` trait to convert from `ToolInfo` To `ConfigAsset`.

### Additional tasks

- [x] Tests added

### DIFF
--- a/src/config/cli.rs
+++ b/src/config/cli.rs
@@ -20,4 +20,7 @@ pub enum Command {
 
     /// Generate a default .tools.toml file and prints it to std out
     DefaultConfig,
+
+    /// Install a tool if it is hardcoded into internal database
+    Install { name: String },
 }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -5,15 +5,16 @@ mod model;
 mod sync;
 
 use std::collections::BTreeMap;
+use std::fmt::Write;
 use std::path::PathBuf;
 
 use clap::Parser;
 
 use crate::config::cli::{Cli, Command};
-use crate::config::schema::ConfigAsset;
+use crate::config::schema::{Config, ConfigAsset};
 use crate::config::template;
 use crate::config::toml;
-use crate::sync::db::lookup_tool;
+use crate::sync::db::{build_db, lookup_tool};
 use crate::sync::sync;
 
 const DEFAULT_CONFIG_PATH: &str = ".tool.toml";
@@ -23,6 +24,7 @@ pub fn run() {
     let config_path = resolve_config_path(cli.config.clone());
 
     match cli.command {
+        Command::DefaultConfig => generate_config(),
         Command::Sync => match toml::parse_file(&config_path) {
             Err(e) => {
                 err::abort_with(&format!(
@@ -31,20 +33,43 @@ pub fn run() {
                     e.display()
                 ));
             }
-            Ok(tool) => {
-                sync(tool);
+            Ok(config) => {
+                sync(config);
             }
         },
-        Command::DefaultConfig => generate_config(),
         Command::Install { name } => {
-            if let Some(tool_info) = lookup_tool(&name) {
-                if let Ok(mut tool) = toml::parse_file(&config_path) {
+            with_parsed_file(&config_path, |mut config| {
+                if let Some(tool_info) = lookup_tool(&name) {
                     let tool_btree: BTreeMap<String, ConfigAsset> =
                         BTreeMap::from([(name, tool_info.into())]);
-                    tool.tools = tool_btree;
-                    sync(tool);
-                }
-            }
+                    config.tools = tool_btree;
+                    sync(config);
+                } else {
+                    let mut exit_message: String =
+                        format!("Unknown tool: {}\nSupported tools:\n", name);
+                    for tool in build_db().keys().cloned().collect::<Vec<String>>() {
+                        if let Err(e) = writeln!(exit_message, "\t* {}", tool) {
+                            err::abort_suggest_issue(&format!("{}", e));
+                        };
+                    }
+                    err::abort_with(&exit_message);
+                };
+            });
+        }
+    }
+}
+
+fn with_parsed_file<F: FnOnce(Config)>(config_path: &PathBuf, on_success: F) {
+    match toml::parse_file(config_path) {
+        Ok(config) => {
+            on_success(config);
+        }
+        Err(e) => {
+            err::abort_with(&format!(
+                "Error parsing configuration at path {}: {}",
+                config_path.display(),
+                e.display()
+            ));
         }
     }
 }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -4,12 +4,16 @@ mod infra;
 mod model;
 mod sync;
 
-use clap::Parser;
+use std::collections::BTreeMap;
 use std::path::PathBuf;
 
+use clap::Parser;
+
 use crate::config::cli::{Cli, Command};
+use crate::config::schema::ConfigAsset;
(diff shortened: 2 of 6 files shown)


MitchellBerend:
Im not sure what tests can be added to make sure everything works properly.

src/config/cli.rs
```diff
@@ -20,4 +20,7 @@ pub enum Command {
 
     /// Generate a default .tools.toml file and prints it to std out
     DefaultConfig,
+
+    /// Install a tool if it is hardcoded into internal database
+    Install { name: String },
```
chshersh, line 25: Hmm, how does it exactly work in `clap`? 🤔
Shouldn't it be under the `--name` flag?

In the future, I imagine wanting the ability to specify all the `ConfigAsset` fields via separate flags like `--repo` and `--owner`. But it can be done separately in the future 🙂

src/lib.rs
```diff
@@ -32,6 +36,16 @@ pub fn run() {
             }
         },
         Command::DefaultConfig => generate_config(),
+        Command::Install { name } => {
+            if let Some(tool_info) = lookup_tool(&name) {
```
chshersh, line 40: Let's output an error message mentioning all supported tools when the tool is not found, like the issue suggests:

```
Unknown tool: 'xxx'
Supported tools:
    * bat
    * ripgrep
    * ...
```

The formatting code would be similar to what has been done in the default config generation. But I'm not sure code duplication could be easily avoided so it's okay to duplicate the relevant code 👍🏻

src/lib.rs
```diff
@@ -32,6 +36,16 @@ pub fn run() {
...
+                if let Ok(mut tool) = toml::parse_file(&config_path) {
```
chshersh, line 41: The logic for handling parsing errors is similar in the `sync` command:

```rust
Command::Sync => match toml::parse_file(&config_path) {
            Err(e) => {
                err::abort_with(&format!(
                    "Error parsing configuration at path {}: {}",
                    config_path.display(),
                    e.display()
                ));
            }
```

Would it be possible to create a separate Higher-Order function to remove the code duplication? I haven't used function arguments in Rust before but I imagine something like this:

```rust
fn with_parsed_config<F>(config_path: &PathBuf, on_success: F)
    where F: Fn(Config) -> ()
{
    match toml::parse_file(config_path) {
            Err(e) => {
                err::abort_with(&format!(
                    "Error parsing configuration at path {}: {}",
                    config_path.display(),
                    e.display()
                ));
            }
            Ok(config) => {
                on_success(config);
            }
}
```

src/lib.rs
```diff
@@ -32,6 +36,16 @@ pub fn run() {
...
+                    let tool_btree: BTreeMap<String, ConfigAsset> =
+                        BTreeMap::from([(name, tool_info.into())]);
+                    tool.tools = tool_btree;
+                    sync(tool);
```
chshersh, line 45: It's pretty neat how `sync` could be reused here 🙂 
Feels like a hack tbh 😅 

But it's nice that it works. Also, could you rename `tool` to `config` here and in the `Sync` case?

I guess it's just a leftover from some old refactoring but this value really should be named `config` as it has the `Config` type

src/lib.rs
```diff
@@ -32,6 +36,16 @@ pub fn run() {
...
+                }
+            }
+        }
```
chshersh, line 48: Ideas for future improvements (can be done as separate issues):

- [ ] Update the config to add a new tool (so it later will be updated with `sync`)
- [ ] Support all `ConfigAsset` fields as arguments to the `install` commandj
- [ ] Allow downloading the tool from the parsed config if it's already specified in there

chshersh (changes requested):
Really nice! I'm surprised how easy you implemented this command! 🤯 

I have a few suggestions to improve the code and some ideas for future improvements but this MVP is already good enough to be shipped with the `v0.2.0` release after a few code changes 👍🏻

chshersh:
> Im not sure what tests can be added to make sure everything works properly.

I think we can add similar integration tests in the `ci.yml` by calling the `tool install` command and verifying the tool exists. I have an idea to simplify the testing setup but writing a few lines of YAML should be okay for now 👍🏻

src/config/cli.rs
MitchellBerend: Ah okay, I was under the impression the `install` command would only support baked in tools. I wanted it to be so users can just run `tool install xyz` without adding extra flags. Im not sure how to add extra flags here.

src/lib.rs
MitchellBerend: Good call

src/lib.rs
MitchellBerend: Oh ye Im definitely misusing the `Config` struct here. I tried constructing everything from scratch but i realized very quickly I would have to read the `.tools.toml` anyway for the bin dir and the arch name. 

This way has the added benefit of getting other options like the proxy option proposed in #82 for free once that gets added.

src/lib.rs
MitchellBerend: How does this look?

```bash
cargo run -- --config tools.toml install sasdf
    Finished dev [unoptimized + debuginfo] target(s) in 2.31s
     Running `target/debug/tool --config tools.toml install sasdf`
Aborting 'tool-sync' with error:

    * Unknown tool: sasdf
Supported tools:
	* bat
	* difftastic
	* exa
	* fd
	* ripgrep
	* tool-sync
```

src/lib.rs
```diff
@@ -27,11 +33,44 @@ pub fn run() {
...
+                    err::abort_with(&exit_message);
+                };
+            });
+        }
+    }
+}
+
+fn with_parsed_file<F: FnOnce(Config)>(config_path: &PathBuf, on_success: F) {
+    match toml::parse_file(config_path) {
+        Ok(config) => {
```
MitchellBerend, line 64: What do you think about this syntax? Would you rather have the where clause between the function head and body?
The where clause is more math like and it is also cleaner if the trait bounds explode in numbers. Should this be the standard for the project?

src/config/cli.rs
chshersh: Extra flags are just nice features when people don't want to edit the config manually and can just install tools via the command-line.

We can figure out how to support different flags later under a separate issue. Because the CLI itself for the multiple flags requires thinking and discussion.

src/lib.rs
chshersh: This syntax is fine 🙂 

It's not a big line of code so `where` is probably unnecessary.

I'm coming from the Haskell background where I would just write:

```haskell
withParsedFile :: FilePath -> (Config -> IO ()) -> IO ()
```

So in some sense, the current Rust definition is closer to what I'm used to.

chshersh (approved):
That looks great 👏🏻

src/lib.rs
chshersh: The formatting is fine 👍🏻 
My first thought would be to used four spaces instead of tabs. But I guess tabs work too 🤔

PR #94, opened 2022-09-15
Refactor commands into separate modules

chshersh:
Some refactoring to change a few things:

* Move all commands into separate modules to make the top-level `match` in `lib.rs` concise
* Change the text of the `default-config`
* Refactor formatting of tool names into a separate function
* Rename integration test files to specify whether they are used for the the `sync` or `install` command


src/sync/db.rs
```diff
@@ -110,3 +110,18 @@ pub fn build_db() -> BTreeMap<String, ToolInfo> {
...
+/// # [bat]
+/// # [exa]
+/// ```
+pub fn fmt_tool_names<F: FnMut(&String) -> String>(fmt_tool: F) -> String {
+    build_db()
+        .keys()
+        .map(fmt_tool)
+        .collect::<Vec<String>>()
+        .join("\n")
+}
```
chshersh, line 127: I wanted to keep the previous implementation with `writeln!` but wasn't able to specify the proper function type signature so I went for a simpler implementation. It is probably less efficient but it looks pretty simple and the performance overhead is not noticeable actually. It still works instantaneously.

src/config/template.rs
```diff
@@ -25,22 +25,22 @@ fn config_template() -> String {
...
-#     # asset_name.windows = "x86_64-pc-windows-msvc""###,
+    # owner     = "BurntSushi"  # GitHub repository owner
+    # repo      = "ripgrep"     # GitHub repository name
+    # exe_name  = "rg"          # Executable name inside the asset
+
+    # Uncomment to download a specific version or tag.
+    # Without this tag latest will be used
+    # tag       = "13.0.0"
+
+    # Asset name to download on linux OSes
```
MitchellBerend, line 36: I did not realize these options were not global in the config. I guess it does make sense since not all tools will support musl or libc or windows.

MitchellBerend:
Maybe this is just me but I like having the line start with the `#`, this is obviously preference though.

```bash

# This configuration is automatically generated by tool-sync 0.1.0
#######################################
#
# Installation directory for all the tools:
# store_directory = "$HOME/.local/bin"
#
# tool-sync provides native support for some of the tools without the need to
# configure them. Uncomment all the tools you want to install with a single
# 'tool sync' command:
#
# [bat]
# [difftastic]
# [exa]
# [fd]
# [ripgrep]
# [tool-sync]
#
# You can configure the installation of any tool by specifying corresponding options:
#
# [ripgrep]  # Name of the tool (new or one of the hardcoded to override default settings)
#     owner     = "BurntSushi"  # GitHub repository owner
#     repo      = "ripgrep"     # GitHub repository name
#     exe_name  = "rg"          # Executable name inside the asset

#     Uncomment to download a specific version or tag.
#     Without this tag latest will be used
#     tag       = "13.0.0"

#     Asset name to download on linux OSes
#     asset_name.linux = "x86_64-unknown-linux-musl"

#     Uncomment if you want to install on macOS as well
#     asset_name.macos = "apple-darwin"

#     Uncomment if you want to install on Windows as well
#     asset_name.windows = "x86_64-pc-windows-msvc"
```

src/sync.rs
```diff
@@ -29,13 +35,17 @@ fn no_tools_message() {
...
+[bat]
+[exa]
+[fd]
+[ripgrep]
 
-    [bat]
-    [exa]
-    [fd]
-    [ripgrep]
+### END ###
```
MitchellBerend, line 48: Good change actually, this makes it more clear where the config starts and ends.

src/sync/db.rs
MitchellBerend: Oh right `join` actually makes sense here. I was looking for a way to concatenate strings like in python, turns out it is the same...

chshersh:
> Maybe this is just me but I like having the line start with the #, this is obviously preference though.

I don't might changing the format actually. I don't have strong feelings about one or another. Having all the comments as leading comments makes it look more consistent 👍🏻

src/sync/db.rs
chshersh: It's a bit sad it doesn't work with iterators and I need to `collect` first. There's [`itertools`](https://crates.io/crates/itertools) which implements this. But I didn't want to add an extra crate for a minor convenience.

Moreover, according to [this comment](https://www.reddit.com/r/rust/comments/6q4uqc/comment/dkv5g06/?utm_source=reddit&utm_medium=web2x&context=3), `join` on `Vec` is even faster than from `itertools` because it can allocate the required amount of memory.

> The comment is 5 years old so things could've changed since then

PR #95, opened 2022-09-16
[#87] Use statically linked binaries for linux

### LINKED ISSUE #87: Use statically linked binaries for linux
See the discussion:

* https://github.com/taiki-e/upload-rust-binary-action/issues/29

chshersh:
Resolves #87

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -21,10 +21,13 @@ jobs:
   upload-assets:
     strategy:
       matrix:
-        os:
-          - ubuntu-latest
-          - macos-latest
-          - windows-latest
+        include:
+          - target: x86_64-unknown-linux-musl
+            os: ubuntu-latest
+          - target: x86_64-apple-darwin
+            os: macos-latest
+          - target: x86_64-pc-windows-msvc
+            os: windows-latest
     runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v3


chshersh:
I checked that the binary is indeed statically linked:

```shell
$ ldd tool 
	statically linked
```

chshersh:
The release with assets is here:

* https://github.com/chshersh/tool-sync/releases/tag/untagged-fc8f5f054b33d6cc8c5a

I'll keep it for now (in case anyone wants to test it) but I'll delete it before making an actual `v0.2.0` release (so don't really on it!)

MitchellBerend:
This looks like a good change but why do we need to keep the dynamically linked one?

chshersh:
> This looks like a good change but why do we need to keep the dynamically linked one?

It's a bit awkward situation due to mutual dependency. `tool-sync` is hardcoded in the DB with the dynamically linked asset:

https://github.com/chshersh/tool-sync/blob/65b53bdbb51fcbe59d5c57a2b5bd1d9181ea6fcf/src/sync/db.rs#L86-L96

And it's used in tests:

https://github.com/chshersh/tool-sync/blob/65b53bdbb51fcbe59d5c57a2b5bd1d9181ea6fcf/.github/workflows/ci.yml#L100

Unfortunately, GitHub API doesn't return `draft` releases as `latest`. So I can only update the test **and** hard-coded tool-sync info only if there's a release with `musl` assets.

So the easiest solution would be to just keep both of them and remove the dynamically linked asset for the `v0.3.0` version after the `v0.2.0` release goes live.

MitchellBerend (approved):

PR #97, opened 2022-09-18
[#83] Create 'AssetError' and refactor 'select_asset' function to return it

### LINKED ISSUE #83: Create 'AssetError' and test 'select_asset' function
This function selects an asset to download based on config:

https://github.com/chshersh/tool-sync/blob/b79eeb91cfdc3a122e6693d503117f68ff1fb44e/src/model/tool.rs#L72-L86

Currently, it returns `Result<Asset, String>` but the goal is to return a custom type for error: `Result<Asset, AssetError`.

The idea is to replace `String` a custom constructor and add tests for these cases.

- [ ] Create a new enum with two constructors
- [ ] Implement the [std::fmt::Display](https://doc.rust-lang.org/std/fmt/trait.Display.html) trait
- [ ] Replace `String` with enum
- [ ] Write tests

zixuanzhang-x:
Resolves #83

Refactored `ToolInfo.select_asset(&assets)` to return a custom `AssetError` error instead of a string and added tests for the function.

The `AssetError` enum has two possible values:
- `NameUnknown`: this is when `AssetName` does not know what the asset name is for a specific OS the user uses.
- `NotFound`: this is when the `AssetName` is not in the `Assets` slice passed into the `select_asset(&assets)` method.

Implemented `std::fmt::Display` trait for `AssetError` and added tests for it.
 
### Additional tasks

- [x] Tests added

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,9 +104,6 @@ By default `tool-sync` reads configuration from `$HOME/.tool.toml` you can run `
 default-config` to print a default configuration example to std out. You can
 redirect this out put to a file like so `tool default-config >
 $HOME/.tool.toml`.
-TODO: create directory automatically, or prompt user to say yes?
-TODO: tools not added to $PATH automatically
-TODO: colorful output of ERROR
 
 You can also quickly copy the above configuration to the default path by running
 the following command (Unix-only):
--- a/README.md
+++ b/README.md
@@ -103,7 +103,10 @@ store_directory = "~/.local/bin"
 By default `tool-sync` reads configuration from `$HOME/.tool.toml` you can run `tool
 default-config` to print a default configuration example to std out. You can
 redirect this out put to a file like so `tool default-config >
-$HOME/.tools.toml`.
+$HOME/.tool.toml`.
+TODO: create directory automatically, or prompt user to say yes?
+TODO: tools not added to $PATH automatically
+TODO: colorful output of ERROR
 
 You can also quickly copy the above configuration to the default path by running
 the following command (Unix-only):
--- a/src/model/release.rs
+++ b/src/model/release.rs
@@ -18,7 +18,7 @@ pub struct Asset {
 #[derive(Debug, PartialEq, Eq)]
 pub enum AssetError {
     /// Asset name of this OS is unknown
-    NameUnknown,
+    OsSelectorUnknown,
 
     /// Asset name is not in the fetched assets
     NotFound(String),
@@ -27,7 +27,7 @@ pub enum AssetError {
 impl Display for AssetError {
     fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
         match self {
-            Self::NameUnknown => {
+            Self::OsSelectorUnknown => {
                 write!(
                     f,
                     "Unknown asset selector for OS: {}. Specify 'asset_name.your_os' in the cofig.",
--- a/src/model/release.rs
+++ b/src/model/release.rs
@@ -1,4 +1,5 @@
 use serde::Deserialize;
+use std::fmt::{Display, Formatter};
 
 #[derive(Deserialize, Debug)]
 pub struct Release {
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -103,7 +103,7 @@ store_directory = "~/.local/bin"
 By default `tool-sync` reads configuration from `$HOME/.tool.toml` you can run `tool
 default-config` to print a default configuration example to std out. You can
 redirect this out put to a file like so `tool default-config >
-$HOME/.tools.toml`.
+$HOME/.tool.toml`.
```
chshersh, line 106: Thanks for the documentation fix! 🙏🏻

src/model/release.rs
```diff
@@ -1,14 +1,52 @@
...
+mod tests {
+    use super::*;
+
+    #[test]
+    fn display_asset_error() {
+        let asset_name = "test_asset";
+        let error_str = AssetError::NotFound(asset_name.to_string()).to_string();
+        assert_ne!(error_str.find(asset_name), None);
+    }
+}
```
chshersh, line 52: Let's remove this test. It doesn't test much so there's no need to maintain it

src/model/release.rs
```diff
@@ -1,14 +1,52 @@
...
+
+impl Display for AssetError {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Self::NameUnknown => {
+                write!(
+                    f,
+                    "Don't know the asset name for this OS: specify it explicitly in the config"
+                )
```
chshersh, line 33: Let's also improve this error message to provide more details for the user (similar to `model/asset_name.rs`)

```suggestion
                write!(
                    f,
                    "Unknown asset selector for OS: {}. Specify 'asset_name.your_os' in the cofig.",
                    env::consts::OS
                )
```

src/model/release.rs
```diff
@@ -1,14 +1,52 @@
...
     pub size: u64,
 }
+
+#[derive(Debug, PartialEq, Eq)]
+pub enum AssetError {
+    /// Asset name of this OS is unknown
+    NameUnknown,
+
+    /// Asset name is not in the fetched assets
+    NotFound(String),
```
chshersh, line 23: `NameUnknown` and `NotFound` sounds too similar in a semantic sense to me. I could switch those two names and enum still would make sense to me (both OS and Asset name can be either unknown or not found).

In this context, `NotFound` is fine but let's rename `NameUnknown` to `OsSelectorUnknown`. Just to give some connection to the fact that it's more related to OS configuration and not the asset name in the list of assets.

src/model/tool.rs
```diff
@@ -107,3 +107,92 @@ pub struct ToolAsset {
     /// GitHub API client that produces the stream for downloading the asset
     pub client: Client,
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
```
chshersh, line 113: Great work on tests! 👏🏻 

Let's also add one more test for the `OsSelectorUknown` constructor 👍🏻

chshersh (changes requested):
Awesome refactoring! 👏🏻 

I left a few comments with improvements. Happy to merge when they are addressed 🙂

src/model/release.rs
zixuanzhang-x: agree, testing error messages is generally brittle

src/model/release.rs
zixuanzhang-x: good idea

src/model/release.rs
zixuanzhang-x: yeah I struggled while naming these two errors, `OsSelectorUnknown` makes sense

src/model/tool.rs
zixuanzhang-x: good catch. added

chshersh (approved):
Looks amazing! 👏🏻 

Thanks a lot for your contribution, @zixuanzhang-x 🤗

PR #99, opened 2022-09-19
[#88] Replace custom display() method with std::fmt::Display trait implementation

### LINKED ISSUE #88: Refactor custom 'display()' functions to use the 'Display' trait
See the discussion in #83 

The idea is to implement the [std::fmt::Display](https://doc.rust-lang.org/std/fmt/trait.Display.html) trait for data types where we already use a custom `display()` function. Specifically:

- [ ] `TomlError::display`
    * https://github.com/chshersh/tool-sync/blob/b79eeb91cfdc3a122e6693d503117f68ff1fb44e/src/config/toml.rs#L16-L24
- [ ] `ToolError::display`
    * https://github.com/chshersh/tool-sync/blob/b79eeb91cfdc3a122e6693d503117f68ff1fb44e/src/model/tool.rs#L20-L29
- [ ] `UnpackError::display`
    * https://github.com/chshersh/tool-sync/blob/b79eeb91cfdc3a122e6693d503117f68ff1fb44e/src/sync/archive.rs#L27-L37

Implementation example:

```rust
enum TestEnum {
    Var1(String),
    Var2(String),
}


impl fmt::Display for TestEnum {
    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
        match self {
            TestEnum::Var1(s) => write!(f, "[VAR1]: {}", s),
            TestEnum::Var2(s) => write!(f, "[VAR2]: {}", s),
        }
    }
}
```

zixuanzhang-x:
Resolves #88

### DIFF
--- a/src/config/template.rs
+++ b/src/config/template.rs
@@ -1,4 +1,4 @@
-/// This file only holds the template that is used to generate a default .tools.toml.
+/// This file only holds the template that is used to generate a default .tool.toml.
 use crate::sync::db;
 
 pub fn generate_default_config() {
--- a/src/config/cli.rs
+++ b/src/config/cli.rs
@@ -18,7 +18,7 @@ pub enum Command {
     /// Sync all tools specified in configuration file
     Sync,
 
-    /// Generate a default .tools.toml file and prints it to std out
+    /// Generate a default .tool.toml file and prints it to std out
     DefaultConfig,
 
     /// Install a tool if it is hardcoded into internal database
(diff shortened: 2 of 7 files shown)


src/config/template.rs
```diff
@@ -1,4 +1,4 @@
-/// This file only holds the template that is used to generate a default .tools.toml.
+/// This file only holds the template that is used to generate a default .tool.toml.
```
chshersh, line 1: Oops, apparently the documentation was wrong in many places 👀 
Thanks again for fixing it!

chshersh (approved):
Looks neat! 👏🏻 

I have nothing to add, this was a pretty good refactoring 🙂

PR #100, opened 2022-09-19
[#98] changed all formatting functions so they accept a type that implements Display

### LINKED ISSUE #98: Refatoring of all functions that print to stdout/err so they accept a generic type that implements the Display trait
The following locations are potential candidates. These should not have a big impact on the rest of the code base, but they allow all the errors and other enums to have the same implementation strategy.

I would love some feedback on this.
- [x] `src/infra/err.rs`
https://github.com/chshersh/tool-sync/blob/65b53bdbb51fcbe59d5c57a2b5bd1d9181ea6fcf/src/infra/err.rs#L3-L13
- [x] `src/infra/err.rs`
https://github.com/chshersh/tool-sync/blob/65b53bdbb51fcbe59d5c57a2b5bd1d9181ea6fcf/src/infra/err.rs#L15-L31
- [x] `src/infra/err.rs`
https://github.com/chshersh/tool-sync/blob/65b53bdbb51fcbe59d5c57a2b5bd1d9181ea6fcf/src/infra/err.rs#L33-L37
- [x] `src/sync/prefetch.rs`
https://github.com/chshersh/tool-sync/blob/65b53bdbb51fcbe59d5c57a2b5bd1d9181ea6fcf/src/sync/prefetch.rs#L39-L42
- [x] `src/sync/prefetch.rs`
https://github.com/chshersh/tool-sync/blob/65b53bdbb51fcbe59d5c57a2b5bd1d9181ea6fcf/src/sync/prefetch.rs#L44-L64
- [x] `src/sync/progress.rs`
https://github.com/chshersh/tool-sync/blob/65b53bdbb51fcbe59d5c57a2b5bd1d9181ea6fcf/src/sync/progress.rs#L87-L93

MitchellBerend:
Resolves #98

This pr converts all functions that accept `&str` to print them out to stdout/err to accept any type that implements `Display` in its stead. This allows these functions to not care about the implementation of the actual print formatting and leaves that up to the implementation of whatever needs to be printed.

### Additional tasks

- [ ] Tests added

### DIFF
--- a/src/config/schema.rs
+++ b/src/config/schema.rs
@@ -64,14 +64,14 @@ impl Config {
         let expanded_store_directory = shellexpand::full(&self.store_directory);
 
         let store_directory = match expanded_store_directory {
-            Err(e) => err::abort_with(&e),
+            Err(e) => err::abort_with(e),
             Ok(cow_path) => PathBuf::from(cow_path.into_owned()),
         };
 
         let has_store_directory = store_directory.as_path().is_dir();
 
         if !has_store_directory {
-            err::abort_with(&format!(
+            err::abort_with(format!(
                 "Specified directory for storing tools doesn't exist: {}",
                 store_directory.display()
             ));
--- a/src/config/schema.rs
+++ b/src/config/schema.rs
@@ -64,7 +64,7 @@ impl Config {
         let expanded_store_directory = shellexpand::full(&self.store_directory);
 
         let store_directory = match expanded_store_directory {
-            Err(e) => err::abort_with(&e.to_string()),
+            Err(e) => err::abort_with(&e),
             Ok(cow_path) => PathBuf::from(cow_path.into_owned()),
         };
 
--- a/src/infra/err.rs
+++ b/src/infra/err.rs
@@ -16,7 +16,7 @@ pub fn abort_with<Message: Display>(err_msg: &Message) -> ! {
 
 /// Print an error message, suggesting opening an issue and exit with code 1
 /// This function can take in any type that implements the [`Display`] trait
-pub fn abort_suggest_issue<Message: Display + ?Sized>(err_msg: &Message) -> ! {
+pub fn abort_suggest_issue<Message: Display>(err_msg: &Message) -> ! {
     eprintln!(
         r#"Aborting 'tool-sync' with error:
 
--- a/src/infra/err.rs
+++ b/src/infra/err.rs
@@ -3,7 +3,7 @@ use std::process;
 
 /// Print an error message and exit with code 1
 /// This function can take in any type that implements the [`Display`] trait
-pub fn abort_with<Message: Display>(err_msg: &Message) -> ! {
+pub fn abort_with<Message: Display>(err_msg: Message) -> ! {
     eprintln!(
         r#"Aborting 'tool-sync' with error:
 
@@ -16,7 +16,7 @@ pub fn abort_with<Message: Display>(err_msg: &Message) -> ! {
 
 /// Print an error message, suggesting opening an issue and exit with code 1
 /// This function can take in any type that implements the [`Display`] trait
-pub fn abort_suggest_issue<Message: Display>(err_msg: &Message) -> ! {
+pub fn abort_suggest_issue<Message: Display>(err_msg: Message) -> ! {
     eprintln!(
         r#"Aborting 'tool-sync' with error:
 
@@ -35,7 +35,7 @@ this issue:
 
 /// Print just the message and exit
 /// This function can take in any type that implements the [`Display`] trait
-pub fn abort<Message: Display>(err_msg: &Message) -> ! {
+pub fn abort<Message: Display>(err_msg: Message) -> ! {
     eprintln!("{}", err_msg);
     process::exit(1);
 }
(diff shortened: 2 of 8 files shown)


MitchellBerend:
Im not sure what tests I could add here, I haven't actually verified that the tests that are run and succeed actually cover this refactor fully. There `.to_string` call in a test is still required for conversion sake.

src/infra/err.rs
```diff
@@ -1,7 +1,9 @@
+use std::fmt::Display;
 use std::process;
 
 /// Print an error message and exit with code 1
-pub fn abort_with(err_msg: &str) -> ! {
+/// This function can take in any type that implements the [`Display`] trait
+pub fn abort_with<Message: Display>(err_msg: &Message) -> ! {
```
chshersh, line 6: I'm curious if we actually need `&` now.

Previously I used `&str` so I can specify static strings without `String::from()` or similar. But I imagine that `Display` is implemented for both the value itself and the reference type.

It also makes more sense to remove `&` in some semantic meaning. We use errors only once — to report them. So they shouldn't be used afterwards and can be safely moved inside error reporting functions.

So I expect that removing `&` from all type signatures near the `Message` and from all use sites will simplify the code further.

src/infra/err.rs
```diff
@@ -13,7 +15,8 @@ pub fn abort_with(err_msg: &str) -> ! {
 }
 
 /// Print an error message, suggesting opening an issue and exit with code 1
-pub fn abort_suggest_issue(err_msg: &str) -> ! {
+/// This function can take in any type that implements the [`Display`] trait
+pub fn abort_suggest_issue<Message: Display + ?Sized>(err_msg: &Message) -> ! {
```
chshersh, line 19: Why do we need `?Sized` here and not in other functions? 🤔

src/sync/prefetch.rs
```diff
@@ -36,12 +38,14 @@ impl PrefetchProgress {
...
-    fn expected_err_msg(&self, tool_name: &str, msg: &str) {
+    /// This method can take in any type that implements the [`Display`] trait
+    fn expected_err_msg<Message: Display>(&self, tool_name: &str, msg: &Message) {
         let tool = format!("{}", style(tool_name).cyan().bold());
         self.pb.println(format!("{} {} {}", ERROR, tool, msg))
     }
 
-    fn unexpected_err_msg(&self, tool_name: &str, msg: &str) {
+    /// This method can take in any type that implements the [`Display`] trait
+    fn unexpected_err_msg<Message: Display>(&self, tool_name: &str, msg: &Message) {
```
chshersh, line 48: I really like how only type-signatures changed but not the implementation of functions and the call sites became simpler as well! 🚀

chshersh (commented):
Looks neat! I really like this refactoring 🤤 

I have a few questions I'd like to know the answer to before merging. But it already looks great 👏🏻

src/infra/err.rs
MitchellBerend: There is an error that passes the message as a `&str`, the compiler complains about this. I could try and box that slice but that means the slice lives on the heap even though its actually already known at compile time.

---
Maybe the `Box` is actually a good thing, since all the output functions always just take a `Display` trait bound. Having a consistent api does sound nice.

src/infra/err.rs
MitchellBerend: For the abort errors this might not make sense actually since the program exits right after.

I'm not 100% sure what the move semantics are here, I know `println` automatically take references to whatever is passed in but the function still moves the error out of the overarching scope. So in my mind the functions that don't exit the program should always take a reference to the error so it does not get dropped after printing it to stdout/err.

src/sync/prefetch.rs
MitchellBerend: Yes this is a perfect application of polymorphism, it does complicate the code a bit for people that aren't used to generic programming, but this cleans up the code a lot in my opinion.

chshersh: Indeed, I like the simplification in this particular case 🙂 

In the end, the generic function is written only once and you need to understand it once. But it's used many times and it can save a bit of boilerplate each time.

src/config/schema.rs
```diff
@@ -64,14 +64,14 @@ impl Config {
         let expanded_store_directory = shellexpand::full(&self.store_directory);
 
         let store_directory = match expanded_store_directory {
-            Err(e) => err::abort_with(&e.to_string()),
+            Err(e) => err::abort_with(e),
```
chshersh, line 67: This is perfect 👨🏻‍🍳 🤌

src/lib.rs
```diff
@@ -37,7 +37,7 @@ fn resolve_config_path(config_path: Option<PathBuf>) -> PathBuf {
                 path
             }
             None => {
-                err::abort_suggest_issue("Unable to find $HOME directory");
+                err::abort_suggest_issue(Box::new("Unable to find $HOME directory"));
```
chshersh, line 40: I tried removing `Box::new` from here and it compiles fine. Could we remove it then?
```suggestion
                err::abort_suggest_issue("Unable to find $HOME directory");
```

chshersh (commented):
Beautiful! Only one last comment 🙂

src/lib.rs
MitchellBerend: Huh weird, I guess maybe the reference change made this possible.

Issue #93, opened 2022-09-15
[RFC] Refactor integration testing


chshersh:
The way tests are implemented currently, it requires lots of manual changes to the CI config. Because of that, the CI configuration is getting bigger, noisy, harder to maintain and error-prone. It also contains lots of boilerplate to support cross-platform tests:

Just look at these 100 lines of YAML 😨 

https://github.com/chshersh/tool-sync/blob/2bb99cf7724980f32e6ab1f970548b436aad99c3/.github/workflows/ci.yml#L43-L141

I propose 

### Option 1: Move to bash scripts

A simple option would be to move those commands into Shell scripts (Bash and PowerShell). This doesn't solve the problem of having lots of boilerplate but at least it makes the CI config less noisy (although at the cost of having more indirections).

**Pros**

✅ Less noise in the CI config
✅ Pretty simple shell scripting and not so much of the code if you're familiar with shell
✅ Testing CLI parsing as well

**Cons**

❌ Requires more polyglot knowledge
❌ Requires lots of manual updates
❌ Boilerplate
❌ Requires separate static analysis like `shellcheck`

### Option 2: Rust integration tests

Alternatively, all these tests can be written as rust integration tests. They can be put into the `tests/` directory and run as `cargo test` (which we already do)

**Pros**

✅ No polyglot env, only Rust for testing
✅ No boilerplate, only one test suite, reusing `lib` utils
✅ Automatically cross-platform
✅ All Rust tooling works for tests automatically

**Cons**

❌ More code
❌ No CLI parsing testing (maybe it can be done separately with `clap`?)

PR #102, opened 2022-09-19
[#54] Adding extra path and added a new error for the case where multiple assets are found

### LINKED ISSUE #54: Support path 'asset_name/bin/<exe_name>' inside assets
This is needed for some tools. Example:

* https://github.com/cli/cli/releases/tag/v2.14.7

All the search paths are specified here:

https://github.com/chshersh/tool-sync/blob/cb339b09238964e8dc2d1a72389b6624c5bb7599/src/sync/archive.rs#L144-L153

To fix this problem:

- [ ] Add one more path search entry
- [ ] Add tests for `gh`

MitchellBerend:
Resolves #54 

This pr add an extra path that gets checked once the binary is downloaded from github. Additionally it also adds an error case where multiple matches are found.

### Additional tasks

- [x] Tests added

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -98,6 +98,7 @@ jobs:
           if [[ ! -x $SYNC_DIR/fd    ]]; then echo "error on: fd";    false; fi
           if [[ ! -x $SYNC_DIR/rg    ]]; then echo "error on: rg";    false; fi
           if [[ ! -x $SYNC_DIR/tool  ]]; then echo "error on: tool";  false; fi
+          if [[ ! -x $SYNC_DIR/gh  ]]; then echo "error on: gh";  false; fi
 
           # disabled because of: https://github.com/alexcrichton/tar-rs/issues/295
           # if [[ ! -x $SYNC_DIR/tokei ]]; then echo "error on: tokei"; false; fi
@@ -138,6 +139,9 @@ jobs:
           }
           if (!(Test-Path $env:SYNC_DIR\tool.exe)) {
               throw 'error on tool.exe'
+          }
+          if (!(Test-Path $env:SYNC_DIR\gh.exe)) {
+              throw 'error on gh.exe'
           }
 
   rustfmt:
--- a/src/model/tool.rs
+++ b/src/model/tool.rs
@@ -198,6 +198,11 @@ mod tests {
                 name: "asset_3".to_string(),
                 size: 77,
             },
+            Asset {
+                id: 3,
+                name: "not a match".to_string(),
+                size: 77,
+            },
         ];
 
         assert_eq!(
--- a/src/model/tool.rs
+++ b/src/model/tool.rs
@@ -87,8 +87,7 @@ impl ToolInfo {
                     // This is safe because there is exactly 1 element
                     1 => Ok(asset.remove(0)),
                     _ => {
-                        let assets: Vec<String> =
-                            asset.iter().map(|item| item.name.clone()).collect();
+                        let assets: Vec<String> = asset.into_iter().map(|item| item.name).collect();
                         Err(AssetError::MultipleFound(assets))
                     }
                 }
@@ -186,17 +185,17 @@ mod tests {
         let assets = vec![
             Asset {
                 id: 1,
-                name: asset_name.to_string(),
+                name: "asset_1".to_string(),
                 size: 10,
             },
             Asset {
                 id: 2,
-                name: asset_name.to_string(),
+                name: "asset_2".to_string(),
                 size: 50,
             },
             Asset {
                 id: 3,
-                name: asset_name.to_string(),
+                name: "asset_3".to_string(),
                 size: 77,
             },
         ];
@@ -204,9 +203,9 @@ mod tests {
         assert_eq!(
             tool_info.select_asset(&assets),
             Err(AssetError::MultipleFound(vec![
-                "asset".into(),
-                "asset".into(),
-                "asset".into()
+                "asset_1".into(),
+                "asset_2".into(),
+                "asset_3".into()
             ]))
         );
     }
--- a/src/model/tool.rs
+++ b/src/model/tool.rs
@@ -76,12 +76,29 @@ impl ToolInfo {
         match self.asset_name.get_name_by_os() {
             None => Err(AssetError::OsSelectorUnknown),
             Some(asset_name) => {
-                let asset = assets.iter().find(|&asset| asset.name.contains(asset_name));
+                let mut asset = assets
+                    .iter()
+                    .filter(|&asset| asset.name.contains(asset_name))
+                    .map(|asset| asset.to_owned())
+                    .collect::<Vec<Asset>>();
+                match asset.len() {
+                    0 => Err(AssetError::NotFound(asset_name.clone())),
 
-                match asset {
-                    None => Err(AssetError::NotFound(asset_name.to_owned())),
-                    Some(asset) => Ok(asset.clone()),
+                    // This unwrap is safe because there is exactly 1 element
+                    1 => Ok(asset.remove(0)),
+                    _ => {
+                        let assets: Vec<String> =
+                            asset.iter().map(|item| item.name.clone()).collect();
+                        Err(AssetError::MultipleFound(assets))
+                    }
                 }
+
+                //match asset {
+                //    None => {
+                //        Err(AssetError::NotFound(asset_name.clone()))
+                //    },
+                //    Some(asset) => Ok(asset.clone()),
+                //}
             }
         }
     }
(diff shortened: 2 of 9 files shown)


MitchellBerend:
The config I tested with and its result.
```bash
$ cat tools.toml
# This config file was generated for version 0.1.0
#
# # tool-sync default configuration file
#####################################################
#
#
store_directory = "$HOME/rust/tool-sync/bin"
#
#
# [bat]
# [difftastic]
# [exa]
# [fd]
[ripgrep]
# [tool-sync]
#
# To add configuration for other tools these are the config options:
[github]
        owner     = "cli" 
        repo      = "cli" 
        exe_name  = "gh"

        # Uncomment to download a specific version or tag.
        # Without this tag latest will be used
        tag       = "v2.14.7"

# Asset name to download on linux OSes
        asset_name.linux = "linux_amd64"
#
# uncomment if you want to install on macOS as well
# asset_name.macos = "apple-darwin"
#
# uncomment if you want to install on Windows as well
# asset_name.windows = "x86_64-pc-windows-msvc"
```


```bash
$ cargo run -- --config tools.toml sync 
    Finished dev [unoptimized + debuginfo] target(s) in 2.40s
     Running `target/debug/tool --config tools.toml sync`
❌  github 
Multiple name matches found for this asset:

	 * gh_2.14.7_linux_amd64.deb
	 * gh_2.14.7_linux_amd64.rpm
	 * gh_2.14.7_linux_amd64.tar.gz

Please add one of these to the config.
🔄  All done!                                                                                                                                                             📦  Estimated total download size: 2.01 MiB
✅  ripgrep 13.0.0   Completed!                                                                                                                                           ✨  Successfully installed 1 tools!
📁  Installation directory: /home/mitchell/rust/tool-sync/bin
```

src/model/tool.rs
```diff
@@ -76,11 +76,21 @@ impl ToolInfo {
         match self.asset_name.get_name_by_os() {
             None => Err(AssetError::OsSelectorUnknown),
             Some(asset_name) => {
-                let asset = assets.iter().find(|&asset| asset.name.contains(asset_name));
+                let mut asset = assets
+                    .iter()
+                    .filter(|&asset| asset.name.contains(asset_name))
+                    .map(|asset| asset.to_owned())
```
chshersh, line 82: I wonder, could we use `into_iter()` instead of `iter()` to avoid cloning values and return owned values from the list?

src/model/tool.rs
```diff
@@ -76,11 +76,21 @@ impl ToolInfo {
...
+                    0 => Err(AssetError::NotFound(asset_name.clone())),
 
-                match asset {
-                    None => Err(AssetError::NotFound(asset_name.to_owned())),
-                    Some(asset) => Ok(asset.clone()),
+                    // This is safe because there is exactly 1 element
+                    1 => Ok(asset.remove(0)),
+                    _ => {
+                        let assets: Vec<String> =
+                            asset.iter().map(|item| item.name.clone()).collect();
```
chshersh, line 91: Same comment about `into_iter()` here as well

src/model/tool.rs
```diff
@@ -157,6 +167,50 @@ mod tests {
...
+                linux: Some(asset_name.to_string()),
+                macos: Some(asset_name.to_string()),
+                windows: Some(asset_name.to_string()),
+            },
+        };
+
+        let assets = vec![
+            Asset {
+                id: 1,
+                name: asset_name.to_string(),
```
chshersh, line 189: Can we change asset names in this test to have unique test names in each element?
Just to test that the function actually returns different assets from every item and not just replicates the same asset name several times. An easy way to increase test coverage without adding new tests, just changing existing ones 🙂 
```suggestion
                name: "asset_1".to_string(),
```

src/sync/prefetch.rs
```diff
@@ -123,16 +124,23 @@ fn prefetch_tool(
...
-                    Err(err) => {
-                        prefetch_progress.unexpected_err_msg(tool_name, &err);
-                        prefetch_progress.update_message(already_completed);
-                        None
-                    }
+                    Err(err) => match err {
+                        AssetError::MultipleFound(_) => {
+                            prefetch_progress.expected_err_msg(tool_name, &err);
+                            prefetch_progress.update_message(already_completed);
+                            None
```
chshersh, line 136: Indeed, this could be an expected error 🤔 
Nice catch!

src/sync/progress.rs
```diff
@@ -26,7 +26,11 @@ impl SyncProgress {
     /// !!! The given `Vec` must be non-empty !!!
     pub fn new(tool_pairs: Vec<ToolPair>) -> SyncProgress {
         // unwrap is safe here because 'new' is called with a non-empty vector
-        let max_tool_size = tool_pairs.iter().map(|tp| tp.name.len()).max().unwrap();
+        let max_tool_size = tool_pairs
+            .iter()
+            .map(|tp| tp.name.len())
+            .max()
+            .unwrap_or(MIN_TAG_SIZE);
```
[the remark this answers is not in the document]
chshersh: Good catch! Indeed, we shouldn't use `unwrap()` at all and I'd love to ban all usages of `unwrap` (ideally with `clippy`). `expect` is always strictly better because it provides a more informative message.

chshersh (changes requested):
This is a helpful message improvement 👍🏻 

As always, a few comments 🙂

src/model/tool.rs
MitchellBerend: This does not work here unfortunately since `assets` is used again on line 90.

src/sync/prefetch.rs
```diff
@@ -123,16 +124,23 @@ fn prefetch_tool(
 
             match client.fetch_release_info() {
                 Err(e) => {
-                    prefetch_progress.unexpected_err_msg(tool_name, &format!("{}", e));
+                    prefetch_progress.unexpected_err_msg(tool_name, &e);
```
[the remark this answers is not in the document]
MitchellBerend: Actually this signature can take ownership. A reference can still be passed in and the compiler just knows its not moved even though the signature says it takes ownership.

MitchellBerend:
@chshersh The hardcoded `gh` binary on mac now downloads the amd64 version, it does not support arm (yes) but this might be possible in the future. The same is true for x86 and arm for linux, should there be an error in the realm of: 

hey your arch is x but this config/hardcoded tool currently downloads y, add <the correct config block> to your .tool.toml

.github/workflows/ci.yml
```diff
@@ -98,6 +98,7 @@ jobs:
           if [[ ! -x $SYNC_DIR/fd    ]]; then echo "error on: fd";    false; fi
           if [[ ! -x $SYNC_DIR/rg    ]]; then echo "error on: rg";    false; fi
           if [[ ! -x $SYNC_DIR/tool  ]]; then echo "error on: tool";  false; fi
+          if [[ ! -x $SYNC_DIR/gh  ]]; then echo "error on: gh";  false; fi
```
chshersh, line 101: Once #93 is done, we won't need to change add new checks for new tools anymore 🤞🏻

src/model/tool.rs
chshersh: Isn't this an error?

On line 90, you extract `name` from all original assets fetched from GitHub. I think, it would be better to show only those assets that match the provided query.

So, I imagine, if you add an asset with non-matching name in the `multiple_asset_found` test, the test would expect this non-matching asset as well. And I think we want to show only matching assets.

Or do I read the code incorrectly? 🤔

chshersh (commented):
One more question in the review.

About `arm64` — agree, let's keep it for future improvements. Currently, `tool-sync` matches only on [`env::consts::OS`](https://doc.rust-lang.org/std/env/consts/constant.OS.html) but when we want to support an arch, we need to handle [`env::consts::ARCH`](https://doc.rust-lang.org/std/env/consts/constant.ARCH.html) as well. This will complicate the logic significantly (especially when we want to support automatic asset name guess) so this would require a proper thinking ahead 🙂

MitchellBerend:
> On line 90, you extract name from all original assets fetched from GitHub. I think, it would be better to show only those assets that match the provided query.

The variable naming is not optimal here.

This part has `asset` contain a `Vec<Asset>` with only matches.
https://github.com/chshersh/tool-sync/blob/265e1f3105291e979c4ccc6704515ea3d66151f8/src/model/tool.rs#L79-L83

This part then pulls out the names of all matches.
https://github.com/chshersh/tool-sync/blob/265e1f3105291e979c4ccc6704515ea3d66151f8/src/model/tool.rs#L90-L91

This should probably be ensured in the test though so I will add an asset that does not match.

chshersh (approved):
Looks terrific 🏆

PR #103, opened 2022-09-20
[#84] Refactor of the way tools are inserted into the hardcoded db

### LINKED ISSUE #84: More elegant definition of hardcoded tools
Currently, when a new tool inserted in a hardcoded database, it's written like this:

```rust
    tools.insert(
        "bat".into(),
        ToolInfo {
            owner: "sharkdp".to_string(),
            repo: "bat".to_string(),
            exe_name: "bat".to_string(),
            asset_name: AssetName {
                linux: Some("x86_64-unknown-linux-musl".to_string()),
                macos: Some("x86_64-apple-darwin".to_string()),
                windows: Some("x86_64-pc-windows-msvc".to_string()),
            },
            tag: ToolInfoTag::Latest,
        },
    );
```

This is quite verbose. I'd love to be able to use just plain strings here like this:

```rust
    tools.insert(
        "bat",
        ToolInfo {
            owner: "sharkdp",
            repo: "bat",
            exe_name: "bat",
            tag: ToolInfoTag::Latest,
            asset_name: AssetName {
                linux: Some("x86_64-unknown-linux-musl"),
                macos: Some("x86_64-apple-darwin"),
                windows: Some("x86_64-pc-windows-msvc"),
            },
        },
    );
```

I tried doing something like this in the past but the code became quite complicated due to the management of all the lifetimes in the type signatures. But maybe there's a good middle-ground, so I want to try again 🤔

MitchellBerend:
Resolves #84 

This pr adds a local macro that simplifies the syntax for inserting a name and `ToolInfo` into the db.

### Additional tasks

- [ ] Tests added

### DIFF
--- a/src/sync/db.rs
+++ b/src/sync/db.rs
@@ -9,6 +9,136 @@ pub fn lookup_tool(tool_name: &str) -> Option<ToolInfo> {
     known_db.remove(tool_name)
 }
 
+macro_rules! insert_tool_into {
+    //trailing commas
+    (
+        $tools:expr,
+        $tool_name:expr,
+        {
+            owner: $owner:expr,
+            repo: $repo:expr,
+            exe_name: $exe_name:expr,
+            asset_name: {
+                linux: $linux:expr,
+                macos: $macos:expr,
+                windows: $windows:expr,
+            },
+            tag: $tag:expr,
+        }
+    ) => {
+        $tools.insert(
+            $tool_name.to_string(),
+            ToolInfo {
+                owner: $owner.to_string(),
+                repo: $repo.to_string(),
+                exe_name: $exe_name.to_string(),
+                asset_name: AssetName {
+                    linux: Some($linux.to_string()),
+                    macos: Some($macos.to_string()),
+                    windows: Some($windows.to_string()),
+                },
+                tag: $tag,
+            },
+        )
+    };
+
+    // first trailing comma
+    (
+        $tools:expr,
+        $tool_name:expr,
+        {
+            owner: $owner:expr,
+            repo: $repo:expr,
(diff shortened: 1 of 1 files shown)


src/sync/db.rs
```diff
@@ -12,100 +12,122 @@ pub fn lookup_tool(tool_name: &str) -> Option<ToolInfo> {
...
+            repo: "exa",
+            exe_name: "exa",
+            asset_name: {
+                linux: "linux-x86_64-musl",
+                macos: "macos-x86_64",
+                windows: "",
             },
             tag: ToolInfoTag::Latest,
         },
     );
```
MitchellBerend, line 59: Does this `""` conversion to `None` make sense? It starts to get really messy when the asset names need to allow for both `Option<&str>` and `&str`

src/sync/db.rs
```diff
@@ -139,3 +161,95 @@ pub fn fmt_tool_names<F: FnMut(&String) -> String>(fmt_tool: F) -> String {
...
+                },
+                tag: $tag,
+            },
+        )
+    };
+}
+
+// Macros normally have to be defined before they get used, this statement makes it usable above
+// its definition as well.
+use insert_tool_into;
```
MitchellBerend, line 255: Macros in rust for some reason are only available after they are defined, this `use` statement makes it so its actually available before its defined too.

MitchellBerend (commented):
@chshersh can you take a quick look at this? The way the `ToolInfo` is now passed to this macro does not look like idiomatic rust to me but my experience with writing macros is very limited.

MitchellBerend:
Im closing this pr since the current proposed implementation does not involve a macro.

PR #113, opened 2022-09-21
[#92] Install a tool from the config file using `tool sync <tool-name>`

### LINKED ISSUE #92: Install tools from config using 'tool sync <tool-name>'
Add the string argument to the `tool sync` command like so:

```shell
tool sync xyz
```

To allow syncing a single tool from the config. Sometimes you just want to update one of the tools to the latest version (or specific tag) and don't want to redownload all the tools but still use the config.

If the tool is not found in the config, suggest using `tool install` command.

zixuanzhang-x:
Resolves #92 

Added an optional <TOOL> argument for the `sync` command to install a single tool from the config file. If no <TOOL> argument is specified, `sync` will install all the tools from the config file.

### Additional tasks

- [ ] Tests added


src/sync.rs
```diff
@@ -53,9 +64,31 @@ For more details, refer to the official documentation:
...
+fn sync_tool_no_check(mut config: Config, tool: String) {
+    let tool_config_asset = (*config.tools.get(&tool).unwrap()).clone();
+
+    let mut tool_btree = BTreeMap::new();
+    tool_btree.insert(tool.clone(), tool_config_asset);
+
+    config.tools = tool_btree;
+
+    sync_from_config_no_check(config);
+}
```
chshersh, line 90: This has some code duplication with the `install` command:

https://github.com/chshersh/tool-sync/blob/df7d4768c9cff2a91616bd9686bc928a52fc8c10/src/install.rs#L17-L20

Could you instead create a function called `sync_single_tool` that will take as arguments a mutable config, a tool name, `ConfigAsset`, and then creates a singleton map from that and runs `sync_from_config_no_check`?

You can then use this function in both places

src/sync.rs
```diff
@@ -17,15 +18,25 @@ use self::prefetch::prefetch;
...
-    } else {
-        sync_from_config_no_check(config)
+        no_tools_message();
+        return;
+    }
+
+    match tool {
+        Some(tool) => {
+            if config.tools.contains_key(&tool) {
+                sync_tool_no_check(config, tool);
```
chshersh, line 34: You can use [`get`](https://doc.rust-lang.org/std/collections/struct.BTreeMap.html#method.get) here and match on the resulting `Option` instead of checking if the value is in the map. This way you'll perform only one lookup in the map instead of two. And you also won't need to do `(*config.tools.get(&tool).unwrap())`

src/sync.rs
```diff
@@ -53,9 +64,31 @@ For more details, refer to the official documentation:
...
 
+fn tool_not_in_config_message(tool: &str) {
+    eprintln!(
+        r#"Tool: {} is not specified in the configuration file.
+
+Consider adding it to the configuration file or use tool install command
+to install it if it's a known tool.
+        "#,
+        tool
+    );
```
chshersh, line 75: Let's improve the message even further 🙂 

```suggestion
        r#"The '{}' tool is not listed in the configuration file: {}.

Add the tool to the configuration file or use the 'tool install' command for installing one of the tools natively supported by 'tool-sync'."#,
        tool,
        path,
    );
```

chshersh (changes requested):
Looks like a great change! 👏🏻 

@zixuanzhang-x Could you try to run `tool sync xyz` with a tool in the config and not in the config and share some screenshots to show how it works?

It would be great to add tests but it might be a bit of work with the current test setup. It should be easier to add tests once #93 is done 🙂 

Also, could you add your changes to the `Unreleased` section in CHANGELOG and update the [Usage section](https://github.com/chshersh/tool-sync#usage) in README with an example of syncing a single tool? 🙏🏻

src/sync.rs
zixuanzhang-x: good idea

src/sync.rs
zixuanzhang-x: got it, it does make the code simpler

zixuanzhang-x:
Here're the screenshots of running `tool sync <tool-name>` when the tool is and is not in the config:
Screen Shot 2022-09-21 at 10 59 15 PM
Screen Shot 2022-09-21 at 11 06 18 PM

src/sync.rs
```diff
@@ -7,25 +7,33 @@ mod prefetch;
...
-        no_tools_message()
-    } else {
-        sync_from_config_no_check(config)
+        no_tools_message();
+        return;
+    }
+
+    match tool {
+        Some(tool) => match config.tools.get(&tool) {
+            Some(asset) => sync_single_tool(config.clone(), tool, (*asset).clone()),
```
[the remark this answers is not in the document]
zixuanzhang-x: oh nice, with the `remove` method I don't need `clone()` indeed.

The reason I previously cloned config is that rust complains `cannot move borrowed value` because config is borrowed in the `match` clause.

chshersh (approved):
Amazing job! 🚀

PR #114, opened 2022-09-21
[#84] refactor of hardcoded tools

### LINKED ISSUE #84: More elegant definition of hardcoded tools

MitchellBerend:
Resolves #84 

This pr aims to simplify the way tools are added so that there aren't any `into()` calls to convert from `&str` to `String`.

### Additional tasks

- [ ] Tests added
- [x] Remove `Into` or `From` implementation

### DIFF
--- a/src/sync/db.rs
+++ b/src/sync/db.rs
@@ -12,7 +12,7 @@ pub fn lookup_tool(tool_name: &str) -> Option<ToolInfo> {
 }
 
 pub fn build_db() -> BTreeMap<String, ToolInfo> {
-    let mut tools = ToolBTreeMap::new();
+    let mut tools = BTreeMap::<&'static str, StaticToolInfo>::new();
 
     tools.insert(
         "bat",
@@ -24,8 +24,7 @@ pub fn build_db() -> BTreeMap<String, ToolInfo> {
             macos: "x86_64-apple-darwin",
             windows: "x86_64-pc-windows-msvc",
             tag: ToolInfoTag::Latest,
-        }
-        .into(),
+        },
     );
     tools.insert(
         "difftastic",
@@ -37,8 +36,7 @@ pub fn build_db() -> BTreeMap<String, ToolInfo> {
             macos: "x86_64-apple-darwin",
             windows: "x86_64-pc-windows-msvc",
             tag: ToolInfoTag::Latest,
-        }
-        .into(),
+        },
     );
     tools.insert(
         "exa",
@@ -50,8 +48,7 @@ pub fn build_db() -> BTreeMap<String, ToolInfo> {
             macos: "macos-x86_64",
             windows: NOT_SUPPORTED,
             tag: ToolInfoTag::Latest,
-        }
-        .into(),
+        },
     );
     tools.insert(
         "fd",
@@ -63,8 +60,7 @@ pub fn build_db() -> BTreeMap<String, ToolInfo> {
             macos: "x86_64-apple-darwin",
             windows: "x86_64-pc-windows-msvc",
             tag: ToolInfoTag::Latest,
-        }
-        .into(),
+        },
     );
     tools.insert(
         "ripgrep",
@@ -76,8 +72,7 @@ pub fn build_db() -> BTreeMap<String, ToolInfo> {
             macos: "apple-darwin",
             windows: "x86_64-pc-windows-msvc",
             tag: ToolInfoTag::Latest,
-        }
-        .into(),
+        },
     );
     tools.insert(
         "tool-sync",
@@ -89,8 +84,7 @@ pub fn build_db() -> BTreeMap<String, ToolInfo> {
             macos: "x86_64-apple-darwin.tar.gz",
             windows: "x86_64-pc-windows-msvc.zip",
             tag: ToolInfoTag::Latest,
-        }
-        .into(),
+        },
     );
     tools.insert(
         "github",
@@ -102,8 +96,7 @@ pub fn build_db() -> BTreeMap<String, ToolInfo> {
             macos: "macOS_amd64",
             windows: "windows_amd64.zip",
             tag: ToolInfoTag::Latest,
-        }
-        .into(),
+        },
     );
     //tools.insert(
     //    "tokei",
@@ -116,10 +109,13 @@ pub fn build_db() -> BTreeMap<String, ToolInfo> {
     //        windows: "x86_64-pc-windows-msvc",
     //        tag: ToolInfoTag::Latest,
     //    }
-    //    .into(),
     //);
 
-    tools.into()
+    BTreeMap::from_iter(
+        tools
+            .into_iter()
+            .map(|(name, tool_info)| (name.to_owned(), tool_info.into())),
+    )
 }
 
 /// Format tool names of the database using a mutating formatting function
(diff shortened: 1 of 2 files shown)


src/sync/db.rs
```diff
@@ -132,10 +129,74 @@ pub fn build_db() -> BTreeMap<String, ToolInfo> {
...
 /// # [exa]
 /// ```
-pub fn fmt_tool_names<F: FnMut(&String) -> String>(fmt_tool: F) -> String {
+pub fn fmt_tool_names<F: FnMut(&'static str) -> String>(fmt_tool: F) -> String {
     build_db()
         .keys()
         .map(fmt_tool)
         .collect::<Vec<String>>()
         .join("\n")
 }
```
MitchellBerend, line 138: @chshersh can you take a look at this `fmt_tool_names` function? It takes a closure and I tried changing the argument to `&str` but the compiler is giving some type errors and I can't figure out how to resolve them.

MitchellBerend: Nvm I think wrapping the `std::collections::BTreeMap` in an internal one with another `From` implementation to convert between the 2 works fine.

src/sync/db.rs
```diff
@@ -3,126 +3,123 @@ use std::collections::BTreeMap;
...
-                windows: Some("x86_64-pc-windows-msvc".to_string()),
-            },
+        "bat",
+        StaticToolInfo {
+            owner: "sharkdp",
+            repo: "bat",
+            exe_name: "bat",
+            linux: "x86_64-unknown-linux-musl",
+            macos: "x86_64-apple-darwin",
+            windows: "x86_64-pc-windows-msvc",
```
chshersh, line 25: This looks so much cleaner and nicer 🤌 

Maybe, one day we'll move all the tools to a TOML file. But I'm not sure about the best way to distribute it and I don't want to download it on start and store somewhere. So... This should be fine for now 🆗

src/sync/db.rs
```diff
@@ -139,3 +136,68 @@ pub fn fmt_tool_names<F: FnMut(&String) -> String>(fmt_tool: F) -> String {
...
+    pub tag: ToolInfoTag,
+
+    pub linux: &'static str,
+    pub macos: &'static str,
+    pub windows: &'static str,
+}
+
+impl From<StaticToolInfo> for ToolInfo {
+    fn from(static_tool_info: StaticToolInfo) -> Self {
+        Self {
```
chshersh, line 182: Can we use the `ToolInfo` constructor here for more explicitness?
```suggestion
        ToolInfo {
```

src/sync/db.rs
```diff
@@ -139,3 +136,68 @@ pub fn fmt_tool_names<F: FnMut(&String) -> String>(fmt_tool: F) -> String {
...
+            owner: static_tool_info.owner.to_string(),
+            repo: static_tool_info.repo.to_string(),
+            exe_name: static_tool_info.exe_name.to_string(),
+            asset_name: AssetName {
+                linux: from_supported_asset(static_tool_info.linux.to_string()),
+                macos: from_supported_asset(static_tool_info.macos.to_string()),
+                windows: from_supported_asset(static_tool_info.windows.to_string()),
```
chshersh, line 189: Can we make `from_supported_asset` accept `&'static str` and remove `to_string()` from here?

It'll stil probably remain in `Some(asset_name.to_string())`. But I think the conversion function will be cleaner. And it could potentially safe some performance when the value is `NOT_SUPPORTED` and we can avoid redundant conversion

src/sync/db.rs
```diff
@@ -139,3 +136,68 @@ pub fn fmt_tool_names<F: FnMut(&String) -> String>(fmt_tool: F) -> String {
...
+    fn insert(&mut self, key: &'static str, value: ToolInfo) -> Option<ToolInfo> {
+        self._inner_map.insert(key.into(), value)
+    }
+}
+
+impl From<ToolBTreeMap> for BTreeMap<String, ToolInfo> {
+    fn from(tool_btree: ToolBTreeMap) -> Self {
+        tool_btree._inner_map
+    }
+}
```
chshersh, line 160: A separate wrapper data type for `BTreeMap` feels a bit clunky. Let's use a different (more functional) approach.

In line 15, we can create:

```rust
let mut tools: BTreeMap<&'static str, StaticToolInfo> = BTreeMap::new();
```

And then, we can return:

```rust
BTreeMap::from_iter(
    tools
        .into_iter()
        .map(|(k, v)| (k.to_owned(), v.into()))
)
```

We'll just convert one `BTreeMap` to another by mapping both key and value.

<hr>

I tried playing a bit with `&'static str` but also had some weird errors with `FnMut`. So, let's try a simpler approach. I'm still not fluent with Higher-Order Functions in the Rust sense 🥴

chshersh (changes requested):
This looks super neat! 💎 

I have a few ideas about simplifying the code but I like how much nicer it looks now 🤤

src/sync/db.rs
MitchellBerend: I will change the return type as well. I always return `Self` so if the name of the struct changes it's that many less places that also need  change.

src/sync/db.rs
MitchellBerend: This is actually a way simpler approach. I tried looking for a solution in the `fmt_tool_names` function as well but it got messy quite quickly.

src/sync/db.rs
MitchellBerend: Ah yes there might be some `""` that get converted to `String` for no reason. I initially had them here so the `From` trait would do all the conversion, but this is a good reason to move the conversion.

src/sync/db.rs
```diff
@@ -3,126 +3,123 @@ use std::collections::BTreeMap;
...
+            owner: "sharkdp",
+            repo: "bat",
+            exe_name: "bat",
+            linux: "x86_64-unknown-linux-musl",
+            macos: "x86_64-apple-darwin",
+            windows: "x86_64-pc-windows-msvc",
             tag: ToolInfoTag::Latest,
-        },
+        }
+        .into(),
```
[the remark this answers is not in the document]
MitchellBerend: This `into()` call is now moved to the `map` at the end of the function. Imo the way tools are added now is very clean.

Issue #50, opened 2022-08-29
Support full automatic asset guess even when 'AssetName' is not provided


chshersh:
CLI tools on GitHub releases follow common naming schemes. `tool-sync` could guess asset names from releases without the need to specify part of the name.

chshersh:
I propose the following design for better maintainability and understandability:

**Model selection rules as an enum data type in a declarative way and specify rules for each OS.**

To be more precise, model query as the following type:

```rust
enum Rule {
    /// Asset name should contain a specified substring
    Substring(String),
    /// Asset name shouldn't contain a specified substring
    Exclude(String),
    /// Both rules should be satisfied
    And(Box<Rule>, Box<Rule>),
    /// At least one rule should be satisfied
    Or(Box<Rule>, Box<Rule>),
}
```

Then, one can easily implement a recursive function for matching a string with a rule:

```rust
fn match(rule: Rule, s: &str) -> bool { ... }
```

A possible rule for macOS could look like this with some eDSL capabilities (after implementing `BitAnd` and `BitOr` traits):

```rust
Substring("apple-darwin") & (Substring(".zip") | Substring(".tar.gz")) & Exclude(".sha256") & Exclude(".md5")
```

Then, we just define such rules for each OS. And, ideally, users of `tool-sync` should never need to specify asset names manually 🤞🏻 

However, users still might want the ability to override the default rules for specific assets in case our rules don't cover their use case or some specific tool. So there's an open question: how much of this eDLS do we want to expose to users? Maybe we should just add another constructor like `Exact` to just let users specify the asset name exactly, without any substrings or regular expression checks

cc @MitchellBerend @slyshykO What do you think?

MitchellBerend:
I like this idea, not operations in regex are a real pain to do. I think the implementation of this feature depends on how much of the functionality we want exposed to the user. I think regex is common enough to assume users already know how to write correct patterns.

PR #117, opened 2022-09-22
[#116] prevent sha from being included in asset name hits

### LINKED ISSUE #116: Can't sync starship
https://github.com/starship/starship/releases

```toml
[starship] 
owner = "starship"
repo = "starship"
exe_name = "starship"
asset_name.windows = "starship-x86_64-pc-windows-msvc.zip"
```

```sh
>tool sync
x  starship
Multiple name matches found for this asset:

         * starship-x86_64-pc-windows-msvc.zip
         * starship-x86_64-pc-windows-msvc.zip.sha256

Please add one of these to the config.
```

MitchellBerend:
Resolves #116 
This pr adds filters out the `sha` substring in the asset name so it does not get trigger a `MultipleFound` error. It also augments a test to check this behaviour.

Thanks to @slyshykO for finding this bug and reporting it.

### Additional tasks

- [x] Tests added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,12 @@ available [on GitHub][2].
 
 ## [Unreleased]
 
+### Fixed
+
+* [#116](https://github.com/chshersh/tool-sync/issues/116):
+  Prevent sha from being included in asset name hits
+  (by [@MitchellBerend][MitchellBerend])
+
 ## [0.2.0] — 2022-09-20 🔃
 
 ### Added
--- a/src/model/tool.rs
+++ b/src/model/tool.rs
@@ -78,7 +78,9 @@ impl ToolInfo {
             Some(asset_name) => {
                 let mut filtered_assets = assets
                     .iter()
-                    .filter(|&asset| asset.name.contains(asset_name) && !asset.name.contains("sha"))
+                    .filter(|&asset| {
+                        asset.name.contains(asset_name) && !asset.name.contains(".sha")
+                    })
                     .map(|asset| asset.to_owned())
                     .collect::<Vec<Asset>>();
                 match filtered_assets.len() {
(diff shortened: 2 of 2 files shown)


MitchellBerend:
Maybe its better to target a specific region in the asset name? Something like after the last instance of `.` so for example `shadow.tar.gz` still gets recognized as a hit.

slyshykO:
I propose to add a filter to settings and use a wildcard for filtering

MitchellBerend:
Do you have suggestions on what this would look like?

slyshykO:
> Do you have suggestions on what this would look like?

```toml
[starship] 
owner = "starship"
repo = "starship"
exe_name = "starship"
asset_name.windows = "starship-x86_64-pc-windows-msvc.zip"

asset_name.exclude = ["*.tar.gz", "*.sha256"]
```

slyshykO:
@MitchellBerend can I add a new dependency https://crates.io/crates/wildmatch?

chshersh:
Hey, @MitchellBerend @slyshykO, could I ask you not to rush the implementation of this? I would like everyone to agree on the final design before we spent a lot of time implementing code that will be thrown away 🙏🏻 

I'll try to provide my thoughts in #50 and I'll mention everyone to request feedback 🙂 

<hr>

As for `wildmatch`: do you mean to use it to implement a separate `asset_name.exclude` field? I'm not sure I want to go that way. And `wildmatch` doesn't support `$` so we can solve this problem without a separate field.

I'd rather spend some time thinking about proper filtering (inclusion/exclusion) mechanism 🤔

slyshykO:
> As for wildmatch: do you mean to use it to implement a separate asset_name.exclude field? I'm not sure I want to go that way.

As an alternative, there is possible to consider an asset's names as wildcards.
As an example: `*x86_64-pc-windows-msvc.zip` .

MitchellBerend:
I'm closing this pr so others won't see a pr linked to the original issue.

PR #120, opened 2022-09-23
[#119] De-pluralize success message when only 1 tool is installed

### LINKED ISSUE #119: Depluralise tools in the success case
It's a very common case when only a single tool is synced/installed:

* `tool install xyz`: install a new single tool
* `tool sync xyz`: sync only one tool from the config
* Only 1 tool is specified in the config (or left after filtering all errors)

But the success message always uses the plural form `tools`.

I propose to use `tool` instead of `tools` when only 1 tool is installed:

https://github.com/chshersh/tool-sync/blob/dd7a006da995285eb925965472d67f87107d8c19/src/sync.rs#L84

zixuanzhang-x:
Resolves #119 

### Additional tasks

- [ ] Tests added
- [x] Updated CHANGELOG.md

### DIFF
--- a/src/sync.rs
+++ b/src/sync.rs
@@ -107,6 +107,10 @@ pub fn sync_from_config_no_check(config: Config) {
         }
     }
 
+    summary_message(installed_tools, store_directory);
+}
+
+fn summary_message(installed_tools: u64, store_directory: PathBuf) {
     eprintln!(
         "{} Successfully installed {} {}!",
         DONE,
--- a/src/sync.rs
+++ b/src/sync.rs
@@ -107,7 +107,16 @@ pub fn sync_from_config_no_check(config: Config) {
         }
     }
 
-    eprintln!("{} Successfully installed {} tools!", DONE, installed_tools);
+    eprintln!(
+        "{} Successfully installed {} {}!",
+        DONE,
+        installed_tools,
+        if installed_tools == 1 {
+            "tool"
+        } else {
+            "tools"
+        }
+    );
     eprintln!(
         "{} Installation directory: {}",
         DIRECTORY,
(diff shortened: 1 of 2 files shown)


src/sync.rs
```diff
@@ -107,7 +107,16 @@ pub fn sync_from_config_no_check(config: Config) {
...
+    eprintln!(
+        "{} Successfully installed {} {}!",
+        DONE,
+        installed_tools,
+        if installed_tools == 1 {
+            "tool"
+        } else {
+            "tools"
+        }
+    );
```
chshersh, line 119: Let's move this big eprintln and the next one into a separate function to keep the main steps on a higher level. I feel like it'll help with clarity if it'll be something like:

```rust
summary_message(installed_tools, store_directory);
```

Instead of two verbose `eprintln!` calls

src/sync.rs
```diff
@@ -107,7 +107,16 @@ pub fn sync_from_config_no_check(config: Config) {
...
-    eprintln!("{} Successfully installed {} tools!", DONE, installed_tools);
+    eprintln!(
+        "{} Successfully installed {} {}!",
+        DONE,
+        installed_tools,
+        if installed_tools == 1 {
+            "tool"
+        } else {
+            "tools"
+        }
```
chshersh, line 118: You could've used `tool{}` and return either empty string or `"s"` but I actually like this way more as it writes the whole word completely and you can see the difference clearly 🙂

chshersh (commented):
Looks neat and pretty simple! Just one comment to improve the code even further but otherwise looks good 👍🏻

src/sync.rs
zixuanzhang-x: good suggestion! It's certainly more readable after breaking down the large function.

PR #121, opened 2022-09-27
[#82] Adds proxy as a flag and config option

### LINKED ISSUE #82: does not work behind a proxy (windows)

MitchellBerend:
Resolves #82 

This pr adds a proxy flag and a proxy option to the `tool.toml`, if both are supplied the proxy flag supersedes the config option.

### Additional tasks

- [ ] Tests added
- [x] Updated CHANGELOG.md


MitchellBerend:
@chshersh @maku can you both check this out? This implementation functionally works for me, but I don't see a way to actually test this.

src/config/schema.rs
```diff
@@ -38,6 +39,9 @@ pub struct ConfigAsset {
 
     /// Name of the specific asset to download
     pub asset_name: AssetName,
+
+    /// Proxy which will get used for all communication
+    pub proxy: Option<ureq::Proxy>,
```
chshersh, line 44: Is there a particular reason we may need a different proxy for individual assets?

src/config/toml.rs
```diff
@@ -40,54 +44,73 @@ pub fn with_parsed_file<F: FnOnce(Config)>(config_path: PathBuf, on_success: F)
...
-fn decode_config(toml: Value) -> Option<Config> {
+fn decode_config(toml: Value, proxy: Option<String>) -> Option<Config> {
     let str_store_directory = toml.get("store_directory")?.as_str()?;
+    let proxy = match proxy {
+        Some(p) => Some(p),
+        None => match toml.get("proxy") {
+            Some(p) => Some(p.as_str()?.into()),
+            None => None,
+        },
+    };
```
chshersh, line 71: Looks like we can implement it easier using `.or_else()`

src/sync/install.rs
```diff
@@ -82,7 +82,10 @@ impl<'a> Installer<'a> {
         match archive {
             None => Err(format!("Unsupported asset type: {}", tool_asset.asset.name).into()),
             Some(archive) => match archive.unpack() {
-                Err(unpack_err) => Err(unpack_err.to_string().into()),
+                Err(unpack_err) => {
+                    println!("{:?}", &tool_asset);
```
chshersh, line 86: Leftover debug print?

chshersh (commented):
I don't use a proxy so I can't verify whether this works or not 😞 

If anyone who uses `proxy` can test if this really does the job, it would be much appreciated!

cc @maku

src/config/schema.rs
MitchellBerend: https://github.com/chshersh/tool-sync/blob/2b752bc1ec88672fde210bffef66f86800b7f601/src/sync/prefetch.rs#L76-L101

This method makes the first call to the github api. It only has access to the names and `ConfigAsset` struct.

The implementation of this was kind of messy so there could be a better way to implement this so the `Client` gets recycled for each api call instead of a new instance being created for every tool.

chshersh:
Thanks for working on this, @MitchellBerend! 👏🏻 

Just a heads-up, I have more suggestions about changing the code. But before putting more work into this, I'd like to have a confirmation that the feature works. I don't use any proxy and once this feature is merged, it will probably remain in the code forever and require maintenance.

Spending more of my free time maintaining features I don't even use is not something I'm eager to do 😌 
And it's even worse if nobody else is going to use them either 😢

maku:
@MitchellBerend @chshersh sorry for delay, recently I worked mostly remote where I do not have a proxy server. But today I tried the implementation (I cloned https://github.com/MitchellBerend/tool-sync).

chshersh:
@maku Thanks for checking the implementation!

In that case, I'm happy to accept the patch 🙂 I'll get around providing more feedback on the implementation once I have time 👌🏻

MitchellBerend:
@maku what os are you on? I would also like to know if this works on windows.

Should auth not be supported on this? My use case also does not require auth so I did not implement it. Support for this can always be added later when needed so maybe this is work that doesn't need to be done right now.

maku:
@MitchellBerend I tried it on windows

chshersh (approved):
Okay, I'm happy to merge it because it works 👍🏻 

I'll create a separate issue to improve the implementation

chshersh:
And here is the issue with the future improvements suggestions:

* https://github.com/chshersh/tool-sync/issues/130

PR #122, opened 2022-10-01
Provide more detailed TOML decoding errors

### LINKED ISSUE #25: Provide more detailed TOML decoding errors

DukeManh:
Fixes #25 

Provide more detailed TOML decoding errors by adding `DecodeError` enum to represent different error types.

Author's questions:

> Does the toml crate provide better decoding than as_str() and friends? All such functions return Option<...> while it would be better to have Result<..., ...>

I couldn't find better decoding than as_str() but I did map `Option<>` to `Result<_, DecodeError>` so I guess dealing with Option is fine?

### Additional tasks

- [x] Tests added
- [ ] Updated CHANGELOG.md

### DIFF
--- a/src/config/toml.rs
+++ b/src/config/toml.rs
@@ -9,11 +9,23 @@ use crate::infra::err;
 use crate::model::asset_name::AssetName;
 use crate::model::os::OS;
 
-#[derive(Debug, PartialEq, Eq)]
+#[derive(Debug, PartialEq)]
 pub enum TomlError {
     IO(String),
     Parse(toml::de::Error),
-    Decode,
+    Decode(DecodeError),
+}
+
+#[derive(Debug, PartialEq)]
+pub enum DecodeError {
+    MissingKey {
+        key: String,
+    },
+    InvalidType {
+        key: String,
+        expected: Value,
+        found: Value,
+    },
 }
 
 impl Display for TomlError {
@@ -21,7 +33,29 @@ impl Display for TomlError {
         match self {
             TomlError::IO(e) => write!(f, "[IO Error] {}", e),
             TomlError::Parse(e) => write!(f, "[Parsing Error] {}", e),
-            TomlError::Decode => write!(f, "[Decode Error]"),
+            TomlError::Decode(e) => write!(f, "[Decode Error] {}", e),
+        }
+    }
+}
+
+impl Display for DecodeError {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        match self {
+            DecodeError::MissingKey { key } => {
+                write!(f, "The key '{}' is missing (a typo or not specified?)", key)
+            }
+            DecodeError::InvalidType {
+                key,
+                expected,
+                found,
+            } => write!(
+                f,
+                "unexpected value type '{}={}': expected `{}`, found `{}`",
+                key,
+                found,
+                expected.type_str(),
+                found.type_str()
+            ),
         }
     }
 }
@@ -55,30 +89,44 @@ fn parse_string(contents: &str, proxy: Option<String>) -> Result<Config, TomlErr
     contents
         .parse::<Value>()
         .map_err(TomlError::Parse)
-        .and_then(|toml| match decode_config(toml, proxy) {
-            None => Err(TomlError::Decode),
-            Some(config) => Ok(config),
-        })
+        .and_then(|toml| decode_config(toml, proxy).map_err(TomlError::Decode))
 }
 
-fn decode_config(toml: Value, proxy: Option<String>) -> Option<Config> {
-    let str_store_directory = toml.get("store_directory")?.as_str()?;
+fn decode_config(toml: Value, proxy: Option<String>) -> Result<Config, DecodeError> {
+    let str_store_directory = toml.get("store_directory");
+
+    let store_directory = str_store_directory.map_or(
+        Err(DecodeError::MissingKey {
+            key: String::from("store_directory"),
+        }),
+        |directory_config| match directory_config {
+            Value::String(directory) => Ok(directory.clone()),
+            other => Err(DecodeError::InvalidType {
+                key: String::from("store_directory"),
+                expected: Value::String("some_value".into()),
+                found: other.clone(),
+            }),
+        },
+    )?;
+
     let proxy: Option<String> = proxy.or_else(|| match toml.get("proxy") {
         Some(p) => Some(p.as_str().unwrap_or("").into()),
         None => None,
     });
 
-    let store_directory = String::from(str_store_directory);
-
     let mut tools = BTreeMap::new();
 
-    for (key, val) in toml.as_table()?.iter() {
+    let table = toml
+        .as_table()
+        .expect("unable to parse config file to a table");
+
+    for (key, val) in table.iter() {
         if let Value::Table(table) = val {
             tools.insert(key.clone(), decode_config_asset(table, &proxy));
         }
     }
 
-    Some(Config {
+    Ok(Config {
         store_directory,
         tools,
         proxy,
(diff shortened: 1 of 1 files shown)


src/config/toml.rs
```diff
@@ -12,15 +12,32 @@ use crate::model::asset_name::AssetName;
...
     IO(String),
     Parse(toml::de::Error),
-    Decode,
+    Decode(DecodeError),
+}
+
+#[derive(Debug, PartialEq, Eq)]
+pub enum DecodeError {
+    NoStoreDirectory,
+    NotATable,
```
DukeManh, line 21: I'm not sure how to trigger this error case yet though it can happen on line 85. I find out that the config file is always a table if we have `store_directory` defined, which is checked before line 85.

chshersh (changes requested):
Hey @DukeManh,

That's a great start! I propose to go one step further and implement the following design:

- [ ] Introduce a separate data type for `TOML` value types:

```rust
enum TomlType {
    String,
    Integer,
    Float,
    Boolean,
    Datetime,
    Array,
    Table,
}
```

- [ ] Implement `DecodeError` as the following enum, listing all possible errors (very similar to yours but with a few changes):

```rust
pub enum DecodeError {
    /// Missing required TOML key
    MissingKey { key: String },

    /// Expecting a different type of an existing key
    InvalidType { key: String, expected: TomlType, found: toml::Value },
}
```

- [ ] Then, create a `Result<_, DecodeError>` versions of all standard functions from `toml` (but not like all-all, it's not enough only for those that we use. I believe, it's only `as_str` and `as_table`). So it should look something like this:

```rust
fn parse_str(toml: &toml::Value, key: &str) -> Result<String, DecodeError> { ... }
```

In that case, I believe we should be able to use the `?` operator and return the first found error.

Ideally, we want to return all possible errors at once. But that would be more complicated so let's start with this 😌

DukeManh:
@chshersh That's a great suggestion, I'm on it.

DukeManh:
I have some questions.

> Introduce a separate data type for TOML value types

Why do we need a separate type for toml and not use `toml::value`?
>  we want to return all possible errors at once.

Is it something you have in mind? https://doc.rust-lang.org/rust-by-example/error/multiple_error_types.html

chshersh:
> Why do we need a separate type for toml and not use toml::value?

`toml::Value` also stores data inside. When we decode TOML, we don't expect any specific data, we expect only the shape of Data. So, unfortunately, we can't reuse `toml::Value` here.

> Is it something you have in mind? https://doc.rust-lang.org/rust-by-example/error/multiple_error_types.html

No, more like something similar:

* https://www.fpcomplete.com/blog/collect-rust-traverse-haskell-scala/#validation

DukeManh:
@chshersh this is ready, can you review it, please?

src/config/toml.rs
```diff
@@ -8,23 +8,85 @@ use crate::config::schema::{Config, ConfigAsset};
...
+        key: String,
+    },
+    InvalidType {
+        key: String,
+        expected: TomlType,
+        found: Value,
+    },
+}
+
+#[allow(dead_code)]
```
chshersh, line 30: Just curious why this line is needed? 🤔

src/config/toml.rs
```diff
@@ -50,25 +112,38 @@ fn parse_string(contents: &str) -> Result<Config, TomlError> {
     contents
         .parse::<Value>()
         .map_err(TomlError::Parse)
-        .and_then(|toml| match decode_config(toml) {
-            None => Err(TomlError::Decode),
-            Some(config) => Ok(config),
-        })
+        .and_then(|toml| decode_config(toml).map_err(TomlError::Decode))
```
chshersh, line 115: That's a neat simplification enabled by the improvement! 💯

src/config/toml.rs
```diff
@@ -8,23 +8,85 @@ use crate::config::schema::{Config, ConfigAsset};
...
+                key,
+                expected,
+                found,
+            } => write!(
+                f,
+                "unexpected value type '{}={}': expected `{}`, found `{}`",
+                key,
+                found,
+                expected,
+                found.type_str()
```
chshersh, line 68: Wow, `type_str` is a neat function! I didn't know it exists and it's actually the idea behind the `TomlType` enum.

src/config/toml.rs
```diff
@@ -8,23 +8,85 @@ use crate::config::schema::{Config, ConfigAsset};
...
+impl Display for TomlType {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        let name = match self {
+            TomlType::String => "string",
+            TomlType::Integer => "integer",
+            TomlType::Float => "float",
+            TomlType::Boolean => "boolean",
+            TomlType::DateTime => "datetime",
+            TomlType::Array => "array",
+            TomlType::Table => "table",
```
chshersh, line 83: Would be nice to reuse `type_str` here and avoid duplication but I don't see a non-ugly way to do this 🤔

src/config/toml.rs
```diff
@@ -50,25 +112,38 @@ fn parse_string(contents: &str) -> Result<Config, TomlError> {
...
+                found: other.clone(),
+            }),
+        },
+    )?;
 
     let mut tools = BTreeMap::new();
 
-    for (key, val) in toml.as_table()?.iter() {
+    // toml is always a table if 'store_directory' is defined above
+    let table = toml.as_table().unwrap();
```
chshersh, line 138: Could we use `expect` with a unique message string here instead of `unwrap`?

src/config/toml.rs
```diff
@@ -8,23 +8,85 @@ use crate::config::schema::{Config, ConfigAsset};
...
+            TomlError::Decode(e) => write!(f, "[Decode Error] {}", e),
+        }
+    }
+}
+
+impl Display for DecodeError {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        match self {
+            DecodeError::MissingKey { key } => {
+                write!(f, "'{}' is not configured", key)
```
chshersh, line 56: Let's change a wording a bit here
```suggestion
                write!(f, "The key '{}' is missing (a typo or not specified?)", key)
```

chshersh (changes requested):
This looks excellent! 👏🏻 
I expected waaaay more changes to support this but it turned out to be quite small 🙂 

I have a few stylistic suggestions and some questions. But overall, the direction is correct 👍🏻

src/config/toml.rs
DukeManh: The compiler warns that I'm only using `TomlType::String` on line 129 so other types are unused. I am just doing this to suppress that. Maybe comment out the used types for now?

src/config/toml.rs
DukeManh: nice, I didn't know it existed

DukeManh:
@chshersh, I decided to reuse `toml::Value` and `type_str` to represent our expected type. What do you think?

chshersh (approved):
I like this implementation! Indeed, it's much simpler 🙂 
Thanks a lot for improving error messages! 👏🏻

PR #123, opened 2022-10-01
Outputs tools URLs in the default-config command

### LINKED ISSUE #111: Output tools URLs in the 'default-config' command
The default configuration contains only tool names:

https://github.com/chshersh/tool-sync/blob/df7d4768c9cff2a91616bd9686bc928a52fc8c10/tests/default-config.toml#L12-L18

But for many people, those names don't mean anything. So I propose to write links to GitHub repos nearby as well. They could serve as a link to documentation and also a quick way to see the tool. So something like this:

```toml
# [bat]        # https://github.com/sharkdp/bat
# [difftastic] # https://github.com/Wilfred/difftastic
# [exa]        # https://github.com/ogham/exa
# [fd]         # https://github.com/sharkdp/fd
# [github]     # https://github.com/cli/cli
# [ripgrep]    # https://github.com/BurntSushi/ripgrep
# [tool-sync]  # https://github.com/chshersh/tool-sync
```

Tool in the config are generated here:

https://github.com/chshersh/tool-sync/blob/df7d4768c9cff2a91616bd9686bc928a52fc8c10/src/config/template.rs#L9

It's pretty easy to add just the URL but it would be nice to have alignment as well

crudiedo:
Resolves #111

This PR adds a new function "fmt_tool_names_info" that accepts a closure with both name and tool info of every record. 

The config_template function calls it to output the tool URL along with applying the alignment based on the longest name length.


src/config/template.rs
```diff
@@ -6,7 +6,21 @@ pub fn generate_default_config() {
 }
 
 fn config_template() -> String {
-    let tools = db::fmt_tool_names(|name| format!("# [{name}]"));
+    let names_length = db::build_db()
+        .keys()
+        .map(|a| a.len())
+        .collect::<Vec<usize>>();
+    let max_name_length = names_length.iter().max().unwrap_or(&0) + 1;
```
MitchellBerend, line 13: a `max` method exists for all iterators, you are collecting the iterator into a `Vec` and then turning it into an iterator again to take its max value.

This would also solve the reference issue I think.

src/sync/db.rs
```diff
@@ -133,6 +133,21 @@ pub fn fmt_tool_names<F: FnMut(&String) -> String>(fmt_tool: F) -> String {
...
+/// # [bat] # https://github.com/ogham/exa
+/// # [exa] # https://github.com/sharkdp/bat
+/// ```
+pub fn fmt_tool_names_info<F: FnMut((&String, &ToolInfo)) -> String>(fmt_tool: F) -> String {
+    build_db()
+        .iter()
+        .map(fmt_tool)
+        .collect::<Vec<String>>()
+        .join("\n")
+}
```
MitchellBerend, line 149: Very good, the function without the comments is only used in one place. I think this would also benefit from your implementation though.

https://github.com/chshersh/tool-sync/blob/2b752bc1ec88672fde210bffef66f86800b7f601/src/install.rs#L7-L29

MitchellBerend (commented):
Hi @crudiedo,

not the reviewer but I thought I would comment on the things I already saw. There is a `.pre-commit-config.yaml` that you can use with [pre-commit](https://pre-commit.com/), it will run cargo ftm, check and clippy every time you commit a file.

MitchellBerend:
Also don't forget to add your change to the `CHANGELOG.md` so this pr isn't missed on the next release by accident!

crudiedo:
@MitchellBerend thank you for your review and help - that's my first pull request, so it got a bit hasty (: 

As suggested, I've updated the install command error output to display the repo URL as well. 

I've also removed the initial fmt_tool_names function - I understand that's a bit risky to do, but as I can see it's not being used anymore.

I really appreciate your help.

CHANGELOG.md
```diff
@@ -14,6 +14,9 @@ available [on GitHub][2].
   Adds the `tool sync <tool-name>` command to install only one tool
   from the configuration file
   (by [@zixuanzhang-x][zixuanzhang-x])
+* [#111](https://github.com/chshersh/tool-sync/issues/111):
+  Adds repo URLs to the output of `default-config` and `install` commands
+  (by [@crudiedo][crudiedo])
```
chshersh, line 19: Thanks for adding a CHANGELOG entry!

Since it's your first contribution, could you also add a URL to your GitHub profile to the `<!-- Contributors -->` section below?

src/install.rs
```diff
@@ -16,7 +16,17 @@ fn install_tool(config: Config, name: String) {
...
+
+        let tools = fmt_tool_names_info(|(name, info)| {
+            format!(
+                "    * {name} {delim:>padding$} https://github.com/{owner}/{repo}",
+                delim = "#",
+                padding = max_name_length - name.len(),
+                owner = info.owner,
+                repo = info.repo,
+            )
+        });
```
chshersh, line 29: I now realise that the only difference in the output of this case is different prefix:

```
    *
```

vs

```
#
```

In that case, I propose to move all this logic into this function and pass only the prefix. So the function will become much simpler and the call sites of this function will be simplified as well.

We can always refactor and add more functions once we need more flexibility 🙂

src/install.rs
```diff
@@ -16,7 +16,17 @@ fn install_tool(config: Config, name: String) {
     if let Some(tool_info) = lookup_tool(&name) {
         sync::sync_single_tool(config, name, tool_info.into());
     } else {
-        let tools = fmt_tool_names(|name| format!("    * {name}"));
+        let max_name_length: usize = build_db().keys().map(|a| a.len()).max().unwrap() + 1;
```
chshersh, line 19: Let's avoid `unwrap()` and use `unwrap_or_default()`. And let's introduce it as a global `const` value e.g. `DEFAULT_TOOL_NAME_LEN`

chshersh (changes requested):
@crudiedo This looks amazing!

I have a few suggestions to simplify the code further 🙂

src/install.rs
crudiedo: Rust doesn't let us do it like 
```rust
const DEFAULT_TOOL_NAME_LEN: usize = build_db().keys().map(|a| a.len()).max().unwrap_or_default() + 1;
```

since we're calling non-const functions. 

We could just set it up to some fixed value, but in that case, we'll have to check it on every hardcoded database update.  

I suggest calculating it inside the fmt_tool_names_info function, plus we'll avoid duplicated "bulid_db()" call in that case.

src/install.rs
crudiedo: Awesome idea! 

One more thing I've noticed is that the default-config command also wraps names in [{name}] (so it shows like `# [bat]`) while install shows it like `* bat`

Let's pass down a lambda (instead of a prefix) that would format the name so we can keep the original format for both commands.

src/install.rs
chshersh: I was thinking more about:

```
const DEFAULT_TOOL_NAME_LEN: usize = 8;
```

So that we don't call unsafe `unwrap` even if we know that the list is not empty but still. It's better to gat this unwrap with some value 🙂

src/install.rs
chshersh: Ah, good catch! Indeed, there's slightly more difference but still possible to refactor 🙂

src/install.rs
crudiedo: Got it :) 

I've replaced `unwrap` with `unwrap_or_default` that should return 0 in case of `None` (since it's `Default` for `usize`)

chshersh: I guess, `unwrap_or_default()` works as well 🤔

CHANGELOG.md
crudiedo: Thanks, done :)

chshersh (approved):
Looks awesome! Thanks a lot 💝

PR #124, opened 2022-10-02
[#110] Introduces the OS enum and improves one error message

### LINKED ISSUE #110: Introduce the 'OS' enum and improve error messages
Currently, `tool-sync` matches the value of `env::consts::OS` with the hardcoded string to extract the relevant part of the asset name:

https://github.com/chshersh/tool-sync/blob/df7d4768c9cff2a91616bd9686bc928a52fc8c10/src/model/asset_name.rs#L18-L22

I propose to introduce an `enum` with 3 constructors. This should allow us to support more OSes in the future if we ever need. And it'll help to improve error messages easily. 

So the issue includes the following tasks:

- [x] Create a new `enum` with 3 OSes
- [x] Create a function to get the current value of enum based on the above matching logic
- [ ] Use `enum` in `get_name_by_os`
- [ ] Implement the `Display` trait for enum to match the TOML configuration format
- [ ] Improve the error message below 

https://github.com/chshersh/tool-sync/blob/df7d4768c9cff2a91616bd9686bc928a52fc8c10/src/model/release.rs#L36-L40

It should be something like:

```rust
"Unknown asset selector for the current OS. Specify 'asset_name.{}' in the config.",
get_os(),
```

crudiedo:
Resolves #110

This PR adds the `OS` enum and updates `AssetName` logic to use it instead of `env::consts::OS`. 

It also improves a single error message to display `Specify 'asset_name.{get_current_os()}'` in it.

### DIFF
--- a/src/config/toml.rs
+++ b/src/config/toml.rs
@@ -7,6 +7,7 @@ use toml::{map::Map, Value};
 use crate::config::schema::{Config, ConfigAsset};
 use crate::infra::err;
 use crate::model::asset_name::AssetName;
+use crate::model::os::OS;
 
 #[derive(Debug, PartialEq, Eq)]
 pub enum TomlError {
@@ -99,9 +100,9 @@ fn decode_asset_name(table: &Map<String, Value>) -> AssetName {
         },
 
         Some(table) => {
-            let linux = str_by_key(table, "linux");
-            let macos = str_by_key(table, "macos");
-            let windows = str_by_key(table, "windows");
+            let linux = str_by_key(table, OS::Linux.to_string().as_str());
+            let macos = str_by_key(table, OS::MacOS.to_string().as_str());
+            let windows = str_by_key(table, OS::Windows.to_string().as_str());
 
             AssetName {
                 linux,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,9 @@ available [on GitHub][2].
 * [#111](https://github.com/chshersh/tool-sync/issues/111):
   Adds repo URLs to the output of `default-config` and `install` commands
   (by [@crudiedo][crudiedo])
+* [#110](https://github.com/chshersh/tool-sync/issues/110):
+  Adds the 'OS' enum and improves error messages
+  (by [@crudiedo][crudiedo])
 
 ### Fixed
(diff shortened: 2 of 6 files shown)


crudiedo:
One specific change I wanted to mention\discuss is this part in `toml.rs` `decode_asset_name`:

```rust
  let linux = str_by_key(table, OS::Linux.to_string().as_str());
  let macos = str_by_key(table, OS::MacOS.to_string().as_str());
  let windows = str_by_key(table, OS::Windows.to_string().as_str());

  AssetName {
      linux,
      macos,
      windows,
  }
```

It wasn't task of the initial issue, but I believe it makes the design a little bit better by directly showing that we're referring to OS-related value and avoiding the situation when we change the OS keys there, but forgot about that error message
```rust
    write!(
        f,
        "Unknown asset selector for the current OS. Specify 'asset_name.{}' in the config.",
        get_current_os()
    )
```

So it'll be pointing to the wrong config key.

src/config/toml.rs
```diff
@@ -99,9 +100,9 @@ fn decode_asset_name(table: &Map<String, Value>) -> AssetName {
         },
 
         Some(table) => {
-            let linux = str_by_key(table, "linux");
-            let macos = str_by_key(table, "macos");
-            let windows = str_by_key(table, "windows");
+            let linux = str_by_key(table, OS::Linux.to_string().as_str());
+            let macos = str_by_key(table, OS::MacOS.to_string().as_str());
+            let windows = str_by_key(table, OS::Windows.to_string().as_str());
```
MitchellBerend, line 106: What do you think about having `&OS::Linux.to_string()` instead of `OS::Linux.to_string().as_str()`?
`&String` gets coerced into `&str` automatically. It makes it more readable for me personally.

```rust
Some(table) => {
            let linux = str_by_key(table, &OS::Linux.to_string());
            let macos = str_by_key(table, &OS::MacOS.to_string());
            let windows = str_by_key(table, &OS::Windows.to_string());
}
```

crudiedo: You're right, it definitely looks more readable.

chshersh (approved):
Thanks a lot for the change! Looks great 👏🏻

Issue #70, opened 2022-09-07
Add checksum check


chshersh:
Adding checksum check (SHA256 or MD5) would enable two features:

* Avoiding the downloading if checksum didn't change
* Verifying the exact matching version

The current problem with supporting this feature is that GitHub doesn't provide this information in their API and maintainers usually don't include the checksum either. One example of a package that has checksum is GitHub CLI:

* https://github.com/cli/cli/releases/tag/v2.15.0

So, first, we need to figure out how to get the reliable source of checksum.

MitchellBerend:
Is there a reason the tool should not just check all the files in the release for a `sha` or `md` sub string? There is a small discussion about requesting the feature [here](https://github.com/orgs/community/discussions/16426). I think we can just support this feature for repos that do publish a checksum.

I would add an optional option in the `tool.toml` like this.

```toml
# This configuration is automatically generated by tool-sync 0.2.0
# https://github.com/chshersh/tool-sync
#######################################
#
# Installation directory for all the tools:
store_directory = "$HOME/rust/tool-sync/bin"
#
# tool-sync provides native support for some of the tools without the need to
# configure them. Uncomment all the tools you want to install with a single
# 'tool sync' command:
#
# [bat]
# [difftastic]
# [exa]
# [fd]
# [github]
# [ripgrep]
# [tool-sync]
#
# You can configure the installation of any tool by specifying corresponding options:
#
[ripgrep]  # Name of the tool (new or one of the hardcoded to override default settings)
    owner     = "BurntSushi"  # GitHub repository owner
    repo      = "ripgrep"     # GitHub repository name
    exe_name  = "rg"          # Executable name inside the asset

    checksum  = "some_checksum_string"
#     Uncomment to download a specific version or tag.
#     Without this tag latest will be used
#     tag       = "13.0.0"

#     Asset name to download on linux OSes
    asset_name.linux = "musl"

#     Uncomment if you want to install on macOS as well
#     asset_name.macos = "apple-darwin"

#     Uncomment if you want to install on Windows as well
#     asset_name.windows = "x86_64-pc-windows-msvc"
```

chshersh:
The `checksum` field is the way to go 👍🏻 
I've checked some common tools. Not many of them provide checksums in their GitHub releases but a few tools to actually provide. And I've added all the tools I've found in the issue body with the description of how they provide the checksum.

Indeed, we can implement this feature by supporting only tools that do provide the checksum. It's not a problem 🙂 

[...]

PR #126, opened 2022-10-04
[#108] Check prefetched `tool_assets` is not empty before passing it to `SyncProgress::new`

### LINKED ISSUE #108: Ensure 'SyncProgress::new()' doesn't take empty Vector
The `SyncProgress::new()` method accepts a vector and requires it to be non-empty:

https://github.com/chshersh/tool-sync/blob/df7d4768c9cff2a91616bd9686bc928a52fc8c10/src/sync/progress.rs#L24-L27

It's created here:

https://github.com/chshersh/tool-sync/blob/df7d4768c9cff2a91616bd9686bc928a52fc8c10/src/sync.rs#L72

However, the `tool_pairs` list might be empty after the `prefetch` phase 💥 

In this case, let's handle this separately and output a message that no tools are required to sync with possible reasons: empty configuration or errors in other tools.

### Additional context

There's the [`nonempty`](https://crates.io/crates/nonempty) crate that provides a non-empty structure. We could use it but it's an extra dependency and for now, we only need it in one place. So let's just ensure the invariant in one single place for now.

zixuanzhang-x:
Resolves #108

### Additional tasks

- [ ] Tests added
- [x] Updated CHANGELOG.md

### DIFF
--- a/src/sync.rs
+++ b/src/sync.rs
@@ -85,7 +85,12 @@ pub fn sync_single_tool(mut config: Config, name: String, asset: ConfigAsset) {
 /// Like `sync_from_config` but expects non-empty list of tools
 pub fn sync_from_config_no_check(config: Config) {
     let store_directory = config.ensure_store_directory();
+
     let tool_assets = prefetch(config.tools);
+    if tool_assets.is_empty() {
+        empty_prefetched_tool_assets_message();
+        return;
+    }
 
     let tool_pairs = tool_assets
         .iter()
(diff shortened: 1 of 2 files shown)


src/sync.rs
```diff
@@ -127,3 +132,10 @@ fn summary_message(installed_tools: u64, store_directory: PathBuf) {
         store_directory.display()
     );
 }
+
+fn empty_prefetched_tool_assets_message() {
+    eprintln!(
+        r"No tools are required to sync: configuration is empty 
+or encountered error prefetching tools."
```
chshersh, line 139: We check if the configuration is empty earlier. So here it can be empty only if all tools have errors.

I suggest changing the `configuration is empty` to `nothing to sync` in case we implement #70 in the future (just to future-proof the message a bit). In that case, the list also will be empty if all versions are up-to-date.

```suggestion
        r"No tools are required to sync: encountered error prefetching tools or nothing to sync."
```

chshersh (commented):
Looks great! I have only one minor nit comment but overall the implementation is what was needed 👍🏻

chshersh (approved):
Thanks!

PR #127, opened 2022-10-04
[#69] Improves error message when the tag is not found

### LINKED ISSUE #69: Improve error messages when the tag is not found
The error message could be improved into saying something like:

```
There's no tag 'latest'
```

or even with the suggestion when possible in some cases:

```
There's no tag 'v13.0.0'. Perhaps you meant '13.0.0'?
```

<hr>

See the following issue for more context:

* https://github.com/chshersh/tool-sync/issues/65

crudiedo:
Resolves #69

This PR improves error message when the tag is not found by adding new `TagError`  error that displays a useful error message. Besides that, the error message may also display a suggestion (Did you mean?) by fetching the available tags and finding the similar one calculating Edit distance. 

### Additional tasks

- [x] Tests added
- [x] Updated CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,9 @@ available [on GitHub][2].
 * [#111](https://github.com/chshersh/tool-sync/issues/111):
   Adds repo URLs to the output of `default-config` and `install` commands
   (by [@crudiedo][crudiedo])
+* [#69](https://github.com/chshersh/tool-sync/issues/69):
+  Adds improved error message with suggesstion when trying to prefetch non-existing tag 
+  (by [@crudiedo][crudiedo])
 
 ### Fixed
 
--- a/src/infra/client.rs
+++ b/src/infra/client.rs
@@ -3,6 +3,7 @@ use std::error::Error;
 use std::io::Read;
 
 use crate::model::release::{Asset, Release};
+use crate::model::tag::Tag;
 
 /// GitHub API client to handle all API requests
 pub struct Client {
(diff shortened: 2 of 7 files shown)


chshersh:
This PR will take a while for me to review due to the implementation complexity and the need to take multiple difficult decisions so I'm lowering its priority in favour of reviewing other PRs.

I've added the label though. Will get back to this PR later when I have enough time.

chshersh:
My proposal is to wait for the implementation of #128 first to see how it affects the design of this issue. As mentioned before, we would like to avoid the total number of GitHub API queries. So a potential solution would be to query GitHub API again for the list of all releases if we don't find the specified tag.

As for the specific implementation of this issue, I propose the following weights for the Levenstein distance algorithm:

* `EDIT_DISTANCE_THRESHOLD = 4` (as current)
* Add weight: 1
* Delete weight: 1
* Replace weight: 2 (aka add + delete; or we can even remove the replace case entirely from the algorithm if it's present and it'll be simulated by add + delete automatically)

The idea behind this is to suggest _similar_ tags. I think it would be weird to suggest version `7.8.9` when the user has specified `1.2.3`. The common case is the absence or presence of the `v` prefix. Or a different minor version (e.g. `13.0.1` vs `13.1.0`).

If there's no release within the edit distance threshold, we could just output all the possible releases with some of their info (e.g. the release date) so the users could select the desired release quickly.

crudiedo:
Hey @chshersh,

I propose to drop that PR in favor of https://github.com/chshersh/tool-sync/pull/135 - even though showing the possible tag is cool, implementing it makes the codebase more complicated. 

The https://github.com/chshersh/tool-sync/pull/135 PR handles both "repo not found" and "tag not found" cases, so it'll be up to the user to check the available tags and choose the one they need.

crudiedo:
Closed due to https://github.com/chshersh/tool-sync/pull/135

PR #135, opened 2022-10-10
[#128] Uses RepoError::NotFound when fetch_release_info returns 404

### LINKED ISSUE #128: Improve error message when the repo is not found
Currently, if we mistype owner/repo in the conf file, we get this error:
```bash
❌  bottom https://api.github.com/repos/clementtsang/notbottom/releases/tags/0.6.3: status code 404
```

It looks a bit confusing for the user since it looks like something tags\releases related. 

I propose to:
- Create new `RepoError::NotFound(owner, repo)` error and implement `fmt::Display` for it (something like "Repo {owner}/repo is not found, please check if it's available")
- Send a `head` request to `https://api.github.com/repos/{owner}/{repo}` before calling `fetch_release_info`
- Catch `ureq::Error::Status(404, _)` on the head request above and create `RepoError` instead 

I believe this would also work for the repos that got deleted.

crudiedo:
Resolves #128

This PR adds a new error `RepoError::NotFound` and uses it when fetch_release_info  returns 404.

The error message depends on the tag, and could be either 
❌  **bottom** The clementtsang/bottomx doesn't exist or tags/0.66.3 was not found.

or

❌  **bottom** The clementtsang/bottomx doesn't exist or has no releases.

if the tag is latest.

### Additional tasks

- [ ] Tests added
- [x] Updated CHANGELOG.md

### DIFF
--- a/src/model/repo.rs
+++ b/src/model/repo.rs
@@ -1,0 +1,28 @@
+use crate::model::tool::LATEST_VERSION;
+use std::fmt::{Display, Formatter};
+
+#[derive(Debug, PartialEq, Eq)]
+pub enum RepoError {
+    /// Either repository or tag is not found due to misconfiguration
+    NotFound {
+        owner: String,
+        repo: String,
+        tag: String,
+    },
+}
+
+impl Display for RepoError {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        match self {
+            RepoError::NotFound { owner, repo, tag } => match tag {
+                _ if tag == LATEST_VERSION => {
+                    write!(f, "The {owner}/{repo} doesn't exist or has no releases.")
+                }
+                _ => write!(
+                    f,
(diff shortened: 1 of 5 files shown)


src/model/repo.rs
```diff
@@ -0,0 +1,28 @@
+use crate::model::tool::LATEST_VERSION;
+use std::fmt::{Display, Formatter};
+
+#[derive(Debug, PartialEq, Eq)]
+pub enum RepoError {
+    /// Either repository or tag is not found due to misconfiguration
+    NotFound {
+        owner: String,
+        repo: String,
+        tag: String,
```
chshersh, line 10: Let's store `ToolInfoTag` here instead of `String` so we can `match` on the constructor instead of a `String` and keep `LATEST_VERSION` internal implementation details

chshersh (commented):
This looks neat and pretty useful! 🏆 

I have one minor refactoring suggestion. But the rest looks good 🙂

chshersh (approved):
Thanks, looks great!

PR #153, opened 2022-10-22
add shellcheck

### LINKED ISSUE #146: groundwork for shellcheck
hello again!

SC has only 1 public release, and well integrated in distro package managers. But it brings up some fun challenges too:

```text
tar tf shellcheck-v0.8.0.linux.x86_64.tar.xz
shellcheck-v0.8.0/LICENSE.txt
shellcheck-v0.8.0/README.txt
shellcheck-v0.8.0/shellcheck
```

- [ ] tar.xz support (and easy addition of `.zst`, `.lzma` as needed)
- [ ] new asset_name guessing pattern (`${project}-v${version}`)

The windows zip puts shellcheck.exe in toplevel, so it works for me right away.

For the dir guessing, i attempted suffix trimming:

```rust
    let platforms = [".x86_64", ".linux"];
    let mut raw_asset = asset_name;
    for p in platforms {
        raw_asset = raw_asset.trim_end_matches(p)
    }
    vec![
        [raw_asset, &exe_name].iter().collect(),
    // .... original guesses
```

but the `platforms` list, as well as other projects' delimiter styles will make this complex fast. What do you think about using `glob` for finding the executable?

---

For xz, i plan to:

1. Use early-return in `Archive::from` to avoid rightward drift, and match more `tar` types
2. switch on the `tar_path` extension to use different types as `tar_decoder`, return error for unsupported compression
3. rename `TarGz` type to be more general `TarBall`

How do you like this?

cheers

hdhoang:
Resolves #146

For unsupported tar compressions, i made a string-based `io::Error`. I tried adding an `AssetError` variant for clarity, but constructing it in `archive.rs` proved much difficult. But it shouldn't be reachable because of `Archive::from` anyhow.

I added an early `$name-$tag` string because threading `ToolAsset` fields through the function added many `&str` parameters, which then get combined at the leaf `exe_paths`.
### Additional tasks

- [x] Tests added
- [x] Updated CHANGELOG.md

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -124,6 +124,7 @@ jobs:
           if [[ ! -x $SYNC_DIR/just      ]]; then echo "error on: just";      false; fi
           if [[ ! -x $SYNC_DIR/k9s       ]]; then echo "error on: k9s";       false; fi
           if [[ ! -x $SYNC_DIR/procs     ]]; then echo "error on: procs";     false; fi
+          if [[ ! -x $SYNC_DIR/shellcheck ]]; then echo "error on: shellcheck; false; fi
 
           # disabled because of: https://github.com/alexcrichton/tar-rs/issues/295
           # if [[ ! -x $SYNC_DIR/tokei ]]; then echo "error on: tokei"; false; fi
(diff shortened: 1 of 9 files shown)


hdhoang:
deko `AnyDecoder` helps greatly, let's try again elsewhere

PR #154, opened 2022-10-23
Adding attributes to catch stray debug prints before merging

### LINKED ISSUE #148: Left over debug println in proxy implementation should be removed
https://github.com/chshersh/tool-sync/blob/395b21c5d39ac4e39b492657954eb283319b2291/src/infra/client.rs#L27-L57

These 2 functions have left over `println` calls from development. These look really weird when syncing a bunch of tools, instead of reusing the first shell line `tool` will print out `None` or the proxy passed in and continue to do that until all info is fetched.

image

### LINKED ISSUE #156: Adding lint options to prevent stray debug println
This is a continuation of #148. The code snippet below can be added at the top of`src/lib.rs` to prevent `println!` that are used for debugging purposes, from making it into the main branch.

```rust
#![deny(clippy::print_stdout)]

#[deny(clippy::print_stdout)]
mod completion;
#[deny(clippy::print_stdout)]
mod config;
#[deny(clippy::print_stdout)]
mod infra;
#[deny(clippy::print_stdout)]
mod install;
#[deny(clippy::print_stdout)]
mod model;
#[deny(clippy::print_stdout)]
mod sync;
```

MitchellBerend:
~~Resolves #148~~
Resolves #156 

~~This pr does not actually resolve any issue but it will prevent stray debug println from being commited to the `main` branch in the future.~~

This pr adds attributes to all the modules imported in `src/lib.rs`. These attributes prevent stray debug `println!`s from making it into the `main` branch. 

For future reference, you can annotate your function or method with the following attribute so clippy passes ci:
```rust
#[allow(clippy::print_stdout)]
fn your_function() {}
```

### Additional tasks

~~- [ ] Documentation for changes provided/changed~~
~~- [ ] Tests added~~
~~- [ ] Updated CHANGELOG.md~~


MitchellBerend:
> @MitchellBerend The issue is about removing extra calls and it's now resolved by @jim4067 in https://github.com/chshersh/tool-sync/pull/152.
Could you open a separate issue about clippy warnings so we don't lose this useful info until we improve our DX? I didn't have the chance to look into it yet disappointed
Also, IIUC, it should be enough to put these clippy options only in src/lib.rs to apply them to the entire tool-sync which is not that bad actually. So maybe we can do this instead of patching our CI config thinking
MitchellBerend reacted with thumbs up emoji

@chshersh You were right with this. This needs to be added at the top of `src/lib.rs`

```rust
#![deny(clippy::print_stdout)]

#[deny(clippy::print_stdout)]
mod completion;
#[deny(clippy::print_stdout)]
mod config;
#[deny(clippy::print_stdout)]
mod infra;
#[deny(clippy::print_stdout)]
mod install;
#[deny(clippy::print_stdout)]
mod model;
#[deny(clippy::print_stdout)]
mod sync;
```

PR #155, opened 2022-10-23
[#150] Added more helpful error message if config toml file not found…

### LINKED ISSUE #150: Output a helpful message when there's no configuration
Currently, when you run `tool sync` and there's no `.tool.toml` file, you see the following message:

```shell
❯ tool sync
Aborting 'tool-sync' with error:

    * Error parsing configuration at path /home/chshersh/.tool.toml: [IO Error] No such file or directory (os error 2)
```

Instead, it would be better to tell more information:

- [ ] How to generate the default configuration file (copy instructions from the documentation)
- [ ] Tell, how you can specify the path of a custom configuration using the `--config` flag

This would require matching on error here:

https://github.com/chshersh/tool-sync/blob/a64ab97025f635472e8526ef362e6dc57752fd52/src/config/toml.rs#L82-L86

gybrish:
… (#150)

While this improves the error message I could not see a good way to fit the change into a test at present, let me know if there are any changes you'd like me to make.

Resolves #150 

- [ ] Tests added
- [ ] Updated CHANGELOG.md

### DIFF
--- a/src/config/toml.rs
+++ b/src/config/toml.rs
@@ -80,7 +80,20 @@ pub fn with_parsed_file<F: FnOnce(Config)>(
 }
 
 fn parse_file(config_path: &PathBuf, proxy: Option<String>) -> Result<Config, TomlError> {
-    let contents = fs::read_to_string(config_path).map_err(|e| TomlError::IO(format!("{}", e)))?;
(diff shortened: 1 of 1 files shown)


gybrish:
Is it just me or is that build failure just a transient error?

PR #160, opened 2023-01-27
db: update k9s artifact name for 0.27.0

hdhoang:
from derailed/k9s#1910

This will break installing older versions, so users must apply different config if they wish.

### Additional tasks

- [ ] Tests added
- [x] Updated CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,7 +34,7 @@ available [on GitHub][2].
 * [#133](https://github.com/chshersh/tool-sync/issues/133)
   Added shell completion
   (by [@MitchellBerend][MitchellBerend])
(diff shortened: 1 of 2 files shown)


chshersh (approved):
Thanks for the fix! 🙏🏻 

The breaking change in K9S is unfortunate. I don't think `amd64` is more standard than `x86_64` as I've seen various combinations. There's no standard when it comes to GitHub Asset names 😞 But I can see the benefit in being more consistent with the K8S ecosystem.

I'm in favour of supporting the new naming scheme by default. Users, who want to install previous versions, can do so by specifying asset name in the config 👌🏻

PR #161, opened 2023-02-09
Support for regex pattern when specifying asset name

LLouice:
Add support for regex pattern when specifying asset name

This feature does not break the previous substring rule

### Additional tasks

- [x] Tests added
- [x] Updated CHANGELOG.md


src/model/release.rs
```diff
@@ -17,11 +18,15 @@ pub struct Asset {
     pub size: u64,
 }
 
-#[derive(Debug, PartialEq, Eq)]
+//#[derive(Debug, PartialEq, Eq)]
```
chshersh: Let's remove this comment if those two instances are not needed 🧹

chshersh (changes requested):
Hmm, initially, I would be against adding regex support to `tool-sync` but seeing how easy it was to implement it, I think I actually changed my mind!

Thanks for the implementation! I have one minor questions. And also, could you update CHANGELOG and docs README by saying that `tool-sync` now supports regular expressions? Thanks!

Everything else looks great 👍🏻

LLouice:
I'm glad you like it.
I added unit tests and documentation (maybe my documentation could be improved).

PR #162, opened 2023-05-20
[#115] Allow self-updates on Windows

### LINKED ISSUE #115: Self update under windows issues.
Is it possible to sync the tool from the tool under windows?
I am getting the error:
```
NO tool-sync v0.2.0   [error] The process cannot access the file because it is being used by another process. (os error 32)
```

FrancisMurillo:
Allow self-updates on Windows by using [self-replace](https://github.com/mitsuhiko/self-replace) crate. 

I was able to test this by building a release from Linux to Windows and the self-update works. I would appreciate if somebody else can test this with a dedicated Windows setup since it would take me some time to setup Windows for Rust. Going to add a Windows CI test if I can.

One caveat with `std::env::current_exe` is that if the executable is a symbolic link, this may be confused with updating the target of the link or the link itself. Given that the executables are in the `store_directory` and that should be in `PATH`, the executables should not be a symbolic link so this should not be an issue. Still this may be an edge case.

Resolves #115

### Additional tasks

- [x] Tests added
- [ ] Updated CHANGELOG.md


FrancisMurillo:
I think the Windows and Unix CI test is sufficient for testing the self-update so I think the PR is ready for review now.

